PR #8, opened 2026-03-20
fix: support bare array format for SUPABASE_JWKS

tomaspozo:
## Summary
- The Supabase platform provides `SUPABASE_JWKS` as a bare JSON array `[{...}]`, but `parseJwks` expected `{ keys: [...] }` format
- Now auto-wraps bare arrays in `{ keys: [...] }` so both formats work
- Adds integration test using realistic platform env var shapes

### DIFF
--- a/src/core/resolve-env.test.ts
+++ b/src/core/resolve-env.test.ts
@@ -99,6 +99,39 @@ describe('resolveEnv', () => {
     })
   })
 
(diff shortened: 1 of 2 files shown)


kallebysantos (approved):
Looks good to me!! 
its similar to this [other commit](https://github.com/supabase/edge-functions-ingress/pull/385/changes/2c86c20cf7adcf3d32f975d4ef0cc8912c8846fd) from ingress

PR #9, opened 2026-03-20
fix: key name resolution for client creation

tomaspozo:
## Summary
- `verifyCredentials` now returns the matched `keyName` in `AuthResult`, so downstream code knows which named key was used for auth
- `createContextClient` and `createAdminClient` use the matched key name instead of always defaulting to `'default'` — with fallback to first available key when no `'default'` exists
- README samples updated to `export default { fetch }` format, removed repetitive imports, and documented bare array JWKS format

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,12 @@ Server-side utilities for Supabase. Handles auth, client creation, and context i
 ```ts
 import { withSupabase } from '@supabase/server'
 
-Deno.serve(
-  withSupabase({ allow: 'user' }, async (req, ctx) => {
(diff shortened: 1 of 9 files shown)


README.md
kallebysantos, line 1: This `export default` examples are not related with the PR context.

kallebysantos (approved):
Looks good to me!!
Just noticing that some docs was changed outside main PR context...

README.md
tomaspozo: ups! sorry, will keep 2 eyes on that in the future

PR #16, opened 2026-03-24
feat: passing down Database generic type to `createClient`

kallebysantos:

feature

Currently there's no way to get Database schema types on `ctx`

## What is the new behavior?

Accepting generics and passing down the database type to `createClient`

## Additional context

<details>

https://github.com/user-attachments/assets/3addb23e-8b50-4aef-8615-bf90825083c8

</details>

### DIFF
--- a/src/core/create-admin-client.ts
+++ b/src/core/create-admin-client.ts
@@ -4,10 +4,10 @@ import { EnvError } from '../errors.js'
 import type { SupabaseEnv } from '../types.js'
 import { resolveEnv } from './resolve-env.js'
 
-export function createAdminClient(
+export function createAdminClient<Database = any>(
   env?: Partial<SupabaseEnv>,
   keyName?: string | null,
-): SupabaseClient {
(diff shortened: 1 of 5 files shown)


src/core/create-admin-client.ts
```diff
@@ -4,10 +4,10 @@ import { EnvError } from '../errors.js'
 import type { SupabaseEnv } from '../types.js'
 import { resolveEnv } from './resolve-env.js'
 
-export function createAdminClient(
+export function createAdminClient<Database = any>(
```
tomaspozo: @kallebysantos we are getting lint errors because of this, can we add a comment to let the linter know is intentional and why?

I see the `supabase-js` client also uses `any`

tomaspozo (approved):

src/core/create-admin-client.ts
kallebysantos: I changed it to `unknown`

PR #19, opened 2026-03-25
feat!: add `supabaseOptions` and refactor client creation to options objects

tomaspozo:
## Summary
- Adds `supabaseOptions?: SupabaseClientOptions<string>` to `WithSupabaseConfig`, allowing users to customize internal `createClient()` calls (e.g., `db.schema`, custom `fetch`, realtime config)
- Options flow through to both the RLS-scoped client and the admin client
- Security-critical auth settings (`persistSession`, `autoRefreshToken`, `detectSessionInUrl`) and the `Authorization` header are always force-overwritten


src/core/create-admin-client.ts
```diff
@@ -42,11 +41,25 @@ export function createAdminClient<Database = unknown>(
...
+      headers: safeHeaders,
+    },
     auth: {
+      ...supabaseOptions?.auth,
       persistSession: false,
       autoRefreshToken: false,
       detectSessionInUrl: false,
     },
-  })
+  } as Parameters<typeof createClient<Database>>[2])
```
kallebysantos, line 64: I believe this ensures TS sync with database main type with the scoped selected schema 🤔

kallebysantos (approved):
Looks good to me!!
Just not really sure if the final Typescript will scope the selected schema, since its expects narrow types and passing down the `SchemaName`.

---

Update: Its not scoping to selected schema
<details>

https://github.com/user-attachments/assets/b98366a3-4922-4de6-91fc-672ed4808439

</details>

src/core/create-admin-client.ts
kallebysantos: Update: It doesn't sync types

PR #22, opened 2026-03-31
feat: exposing `keyName` to `SupabaseContext`

kallebysantos:

Bug fix, feature, docs update

- `SupabaseContext` doesn't have the `auth.keyName` which would be useful for some cases.
- Also, while specifying `secret:key` there's a bug which requires this key to be defined at `publishableKeys` environment prop.

## What is the new behavior?

- Exposing `authKeyName` property to `SupabaseContext`
- Apply `authType` verification before passing down the `keyName` while creating the supabase client


tomaspozo (approved):
This is a great addition! and fix too :)

PR #33, opened 2026-04-21
docs: document userClaims and authKeyName in context

TupiC:

docs update

docs referenced `ctx.user` instead of defined `ctx.userClaims` for JWT-derived claims in user auth mode

## What is the new behavior?

updated docs to correctly reference `ctx.userClaims` and optional `authKeyName`


tomaspozo (approved):
Thank you for your contribution!

PR #46, opened 2026-05-05
feat: add Elysia adapter

wobsoriano:

Feature

`@supabase/server` ships framework adapters for Hono and H3 / Nuxt, but not for Elysia. Users building APIs with Elysia have to wire up Supabase auth manually with the core primitives.

## What is the new behavior?

Adds `@supabase/server/adapters/elysia`, a first-class plugin adapter for Elysia. Elysia has been gaining significant traction in the TypeScript ecosystem, particularly in the Bun community, for its end-to-end type safety and expressive plugin API.

Usage with a plain Elysia app:

```ts
import { Elysia } from 'elysia'
import { withSupabase } from '@supabase/server/adapters/elysia'

const app = new Elysia()
  .use(withSupabase({ auth: 'user' }))
  .get('/games', async ({ supabaseContext }) => {
    const { data: myGames } = await supabaseContext.supabase
      .from('favorite_games')
      .select()

    return myGames
  })

app.listen(3000)
```

Per-route auth using scoped groups:

```ts
const app = new Elysia()
  .get('/health', () => ({ status: 'ok' }))
  .group('/api', (app) =>
    app
      .use(withSupabase({ auth: 'user' }))
      .get('/profile', ({ supabaseContext }) => supabaseContext.userClaims),
  )
```

Custom error handling via Elysia's `onError`:

```ts
const app = new Elysia()
  .use(withSupabase({ auth: 'user' }))
  .onError(({ code, error, status }) => {
    if (code !== 'SupabaseAuthError') return

    const cause = error.cause as { code?: string; status?: number } | undefined

    return status((cause?.status as 401) ?? 500, {
      error: error.message,
      code: cause?.code,
    })
  })
```

## Additional context

- Uses Elysia's `.resolve()` hook with `.as('scoped')` so the context propagates to parent app routes.
- Auth failures throw a registered `SupabaseAuthError`, which integrates with Elysia's standard `onError` flow and preserves the original `AuthError` on `error.cause`.Consumers discriminate on `code === 'SupabaseAuthError'` in `onError` without needing to import the class.
- Test coverage mirrors the Hono and H3 adapter tests.
- Docs and package metadata were updated to include Elysia in the adapter tables, exports, and generated API docs.


src/adapters/elysia/plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+import type { SupabaseContext, WithSupabaseConfig } from '../../types.js'
+
+class SupabaseAuthError extends Error {
+  status: number
+  constructor(message: string, status: number, cause: unknown) {
+    super(message, { cause })
+    this.status = status
+  }
+}
+
```
kallebysantos: I'm not really sure why its `SupabaseAuthError`.
We have [EnvError](https://github.com/supabase/server/blob/ea17d2fdf13064364d33ea996b541b502dfecfca/src/errors.ts#L20) and [AuthError](https://github.com/supabase/server/blob/ea17d2fdf13064364d33ea996b541b502dfecfca/src/errors.ts#L108) classes.

Both sharing similar shape. Maybe better renaming it to `SupabaseError` instead?

src/adapters/README.md
```diff
@@ -35,4 +36,4 @@ The Supabase team will review the PR against these requirements. Once merged, th
 
 ## Designing an adapter
 
-The existing adapters at [`hono/middleware.ts`](hono/middleware.ts) and [`h3/middleware.ts`](h3/middleware.ts) (siblings of this README) are the canonical templates. The shape every adapter exposes is `withSupabase(config, handler)` returning a framework-native middleware. Keep all auth logic in `@supabase/server/core` — adapters should only translate request/response shapes between the framework and the core primitives.
+The existing adapters at [`hono/middleware.ts`](hono/middleware.ts), [`h3/middleware.ts`](h3/middleware.ts), and [`elysia/plugin.ts`](elysia/plugin.ts) (siblings of this README) are the canonical templates. The shape every adapter exposes is `withSupabase(config, handler)` returning a framework-native middleware. Keep all auth logic in `@supabase/server/core` — adapters should only translate request/response shapes between the framework and the core primitives.
```
kallebysantos, line 39: Is "plugin" a name convention for Elysia framework? While "middleware" only applies for the others?

package.json
```diff
@@ -69,24 +73,34 @@
...
+    "elysia": {
+      "optional": true
     }
   },
   "devDependencies": {
     "@commitlint/cli": "^20.4.2",
     "@commitlint/config-conventional": "^20.4.2",
     "@supabase/supabase-js": "^2.98.0",
     "eslint": "^10.0.2",
+    "elysia": "^1.0.0",
```
kallebysantos: is `^1.0.0` correct? all other places references to version `^1.4.0`

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
 
 - **Supabase Edge Functions** — environment variables are auto-injected. Zero config.
 - **Deno / Bun** — works out of the box with the `export default { fetch }` pattern.
-- **Node.js** — use the [Hono adapter](#hono), [H3 adapter](#h3--nuxt), or [core primitives](#primitives) with your framework of choice.
+- **Node.js** — use the [Hono adapter](#hono), [H3 adapter](#h3--nuxt), [Elysia adapter](#elysia), or [core primitives](#primitives) with your framework of choice.
 - **Cloudflare Workers** — enable `nodejs_compat` in `wrangler.toml` or pass env overrides via the `env` config option.
 - **Nuxt** — use the [H3 adapter](#h3--nuxt) directly as a server middleware.
+<<<<<<< HEAD
```
kallebysantos: Please, remove this `git` conflict marks

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
...
 - **Next.js / SvelteKit / Remix** — compose with [`@supabase/ssr`](https://github.com/supabase/ssr): `@supabase/ssr` owns cookies + refresh-token rotation, `@supabase/server` adds verified claims and typed RLS / admin clients on top. See [`docs/ssr-frameworks.md`](docs/ssr-frameworks.md).
 
 ### Does this replace `@supabase/ssr`?
 
 No. `@supabase/ssr` handles cookie-based session management for frameworks like Next.js and SvelteKit. `@supabase/server` handles stateless, header-based auth for Edge Functions, Workers, and other backend runtimes. The composable primitives already work in SSR environments but require more setup — see [`docs/ssr-frameworks.md`](docs/ssr-frameworks.md) for the Next.js example. The two packages coexist and are not replacements for each other. Deeper integration with `@supabase/ssr` is on the roadmap.
+=======
```
kallebysantos: Other conflict mark here

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
...
+- **Elysia** — use the [Elysia adapter](#elysia) as a plugin.
+- **Next.js / SvelteKit / Remix** — use core primitives to build a cookie-based auth adapter. See [`docs/ssr-frameworks.md`](docs/ssr-frameworks.md).
+>>>>>>> 3c2310d (feat: add Elysia adapter)
```
kallebysantos: Another `git` mark

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
...
+| ---------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
+| `@supabase/server`                 | `withSupabase`, `createSupabaseContext`                                                                           |
+| `@supabase/server/core`            | `verifyAuth`, `verifyCredentials`, `extractCredentials`, `createContextClient`, `createAdminClient`, `resolveEnv` |
+| `@supabase/server/adapters/hono`   | `withSupabase` (Hono middleware)                                                                                  |
+| `@supabase/server/adapters/h3`     | `withSupabase` (H3 / Nuxt middleware)                                                                             |
+| `@supabase/server/adapters/elysia` | `withSupabase` (Elysia plugin)                                                                                    |
 
 ## Documentation
 
+<<<<<<< HEAD
```
kallebysantos: Other `git` mark

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
...
 | How do I use this with H3 / Nuxt?                                   | [`docs/adapters/h3.md`](docs/adapters/h3.md)                     |
+| How do I use this with Elysia ?                                     | [`docs/adapters/elysia.md`](docs/adapters/elysia.md)             |
 | How do I use low-level primitives for custom flows?                 | [`docs/core-primitives.md`](docs/core-primitives.md)             |
 | How do environment variables work across runtimes?                  | [`docs/environment-variables.md`](docs/environment-variables.md) |
 | How do I handle errors? What codes exist?                           | [`docs/error-handling.md`](docs/error-handling.md)               |
 | How do I get typed database queries?                                | [`docs/typescript-generics.md`](docs/typescript-generics.md)     |
 | How do I use this with `@supabase/ssr` (Next.js, SvelteKit, Remix)? | [`docs/ssr-frameworks.md`](docs/ssr-frameworks.md)               |
 | What's the complete API surface?                                    | [`docs/api-reference.md`](docs/api-reference.md)                 |
+=======
```
kallebysantos: Here too, another `git` mark

README.md
```diff
@@ -423,39 +465,61 @@ For other environments, pass overrides via the `env` config option or `resolveEn
...
+| How do I use this with Hono?                             | [`docs/hono-adapter.md`](docs/hono-adapter.md)                   |
+| How do I use this with Elysia?                           | [`docs/elysia-adapter.md`](docs/elysia-adapter.md)               |
+| How do I use low-level primitives for custom flows?      | [`docs/core-primitives.md`](docs/core-primitives.md)             |
+| How do environment variables work across runtimes?       | [`docs/environment-variables.md`](docs/environment-variables.md) |
+| How do I handle errors? What codes exist?                | [`docs/error-handling.md`](docs/error-handling.md)               |
+| How do I get typed database queries?                     | [`docs/typescript-generics.md`](docs/typescript-generics.md)     |
+| How do I use this in Next.js, Nuxt, SvelteKit, or Remix? | [`docs/ssr-frameworks.md`](docs/ssr-frameworks.md)               |
+| What's the complete API surface?                         | [`docs/api-reference.md`](docs/api-reference.md)                 |
+>>>>>>> 3c2310d (feat: add Elysia adapter)
```
kallebysantos: And I believe this is the last one, `git` mark

kallebysantos (changes requested):

kallebysantos:
Thanks for contributing, 
Just a few questions and adjustments 💚

src/adapters/elysia/plugin.ts
wobsoriano: Good call! Renamed to `SupabaseError`.

The intent behind a wrapper class rather than throwing `AuthError` directly is to give Elysia consumers a single, stable discriminant in onError regardless of whether the failure was an auth error or an env misconfiguration.

More info https://elysiajs.com/patterns/error-handling

src/adapters/README.md
wobsoriano: Yes, intentional. The idiomatic unit is a "plugin" in elysia.

https://elysiajs.com/essential/plugin.html

package.json
wobsoriano: Fixed, bump to 1.4.0 to match the peer dep range

README.md
wobsoriano: Oof, sorry about those. Removed and cleaned

README.md
wobsoriano: fixed!

wobsoriano:
@kallebysantos thanks for the thorough review. Resolved the comments! ❤️

src/adapters/elysia/plugin.ts
kallebysantos: Uhmm so what about [registering](https://elysiajs.com/patterns/error-handling.html#custom-error) the original error classes instead of wrapping it?

```ts
new Elysia()
  .error({
      EnvError,
      AuthError,
  })
```

package.json
kallebysantos: Seems you need to update the `.lock` file as well

```
Run pnpm install --frozen-lockfile
[ERR_PNPM_OUTDATED_LOCKFILE] Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up to date with <ROOT>/package.json

Note that in CI environments this setting is true by default. If you still need to run install in such cases, use "pnpm install --no-frozen-lockfile"

  Failure reason:
  specifiers in the lockfile don't match specifiers in package.json:
* 1 dependencies are mismatched:
  - elysia (lockfile: ^1.0.0, manifest: ^1.4.0)
```

src/adapters/elysia/plugin.ts
wobsoriano: Good point! I tried throwing `AuthError` directly with `.error({ AuthError })`, but Elysia's error registry inside a scoped plugin doesn't propagate to the parent's `onError` at runtime, so `code === 'AuthError'` silently fails even though TS infers it correctly.

We could keep a thin `SupabaseError` wrapper but expose `.code` and `.status` directly on it (no `.cause` indirection). That way code discrimination still works idiomatically. Open to other suggestions!

wobsoriano: So if we pass the `AuthError` directly, one will check it in the elysia error hook like this:

```ts
import { AuthError } from '@supabase/server/errors'

app.onError(({ error, status }) => {
  if (!(error instanceof AuthError)) return
  return status(error.status as 401, {
    error: error.message,
    code: error.code,
  })
})
```

but if we create a thin wrapper and expose `.code` and `.status`:

```ts
app.onError(({ code, error, status }) => {
  if (code !== 'SupabaseError') return
  return status(error.status as 401, {
    error: error.message,
    code: error.code,
  })
})
```

2nd is more idiomatic to Elysia, no extra import, and the `code` string is the standard discriminant. The first works but bypasses Elysia's custom error system entirely

kallebysantos: > 2nd is more idiomatic to Elysia

Sure! so keep the version you think better aligns to Elysia framework.
Mark it as "resolved" with your confirmation

kallebysantos (approved):
LGTM! 
Please fix the merge conflicts as well give me the confirmation about `SupabaseError` thing.

Then we should be good to merge!

wobsoriano:
> LGTM! Please fix the merge conflicts as well give me the confirmation about `SupabaseError` thing.
> 
> Then we should be good to merge!

Thanks for approving! Going with the thin `SupabaseError` wrapper as it keeps Elysia's idiomatic code discrimination (`code === 'SupabaseError'`) working while exposing `.code` and `.status` directly on the error, so no awkward `.cause` access 🫡

tomaspozo:
> Thanks for approving! Going with the thin `SupabaseError` wrapper as it keeps Elysia's idiomatic code discrimination (`code === 'SupabaseError'`) working while exposing `.code` and `.status` directly on the error, so no awkward `.cause` access 🫡

Hi @wobsoriano! we are ready to merge this... can you confirm it this is already taken care of fully?

wobsoriano:
> > Thanks for approving! Going with the thin `SupabaseError` wrapper as it keeps Elysia's idiomatic code discrimination (`code === 'SupabaseError'`) working while exposing `.code` and `.status` directly on the error, so no awkward `.cause` access 🫡
> 
> Hi @wobsoriano! we are ready to merge this... can you confirm it this is already taken care of fully?

I just pushed an update ([ref](https://github.com/supabase/server/pull/46/changes/58b2cb4aa7ac25d27ced47d736e04dc0f3bbc0a8)) that gives clean access to `.code` and .`status` without a type cast. Also simplified the constructor to take the AuthError directly instead of unpacking its fields.

Tested this locally in a simple Elysia app:

```ts
.onError(({ code, error, status }) => {
  if (code !== 'SupabaseError') return
  return status(error.status {
    error: error.message,
    code: error.cause.code,
  })
})
```

sending a curl request `curl -i http://localhost:3000/ -H "Authorization: Bearer sometoken"` gives

```json
{"error":"Invalid credentials","code":"INVALID_CREDENTIALS"}
```

Ready to merge 👍🏼

tomaspozo:
@wobsoriano 

One small suggestion before merge: would you consider hoisting `code` onto the wrapper too, for consistency with `status`? 

Right now `.status` lives directly on SupabaseError but `.code` still requires going through `.cause`...
  
 ```ts
  class SupabaseError extends Error {
    status: number
    code: string
    declare cause: AuthError
    constructor(inner: AuthError) {
      super(inner.message, { cause: inner })
      this.status = inner.status
      this.code = inner.code
    }
  }
```

Then the handler becomes uniform:

```ts
  .onError(({ code, error, status }) => {
    if (code !== 'SupabaseError') return
    return status(error.status as 401, {
      error: error.message,
      code: error.code,
    })
  })
```

`cause` is still there as the typed AuthError for anyone who wants the full object. Happy to merge as-is if you'd rather keep the wrapper minimal.. Flagging since it's a one-liner 😃

wobsoriano:
@tomaspozo 

Unfortunately it collides with an Elysia internal: when a thrown error has an own `.code` property, Elysia uses its value as the `code` in `onError` instead of the `.error()` registry key. So with `this.code = inner.code`, the handler sees `code === 'INVALID_CREDENTIALS'` instead of `code === 'SupabaseError'`, and the discrimination check never passes.

That's also why `.status` is safe to hoist but `.code` isn't. Elysia doesn't treat `.status` specially. Keeping `.code` off the wrapper and behind `.cause.code` is the workaround.

So if you have this in `onError`:

```ts
.onError(({ code, error }) => {
  console.log('code value:', code)
  console.log('code === SupabaseError:', code === 'SupabaseError')
  
  if (code !== 'SupabaseError') return
  
  // ...
 })
```

The output is:

```
  code value: INVALID_CREDENTIALS
  code === SupabaseError: false
```

tomaspozo:
Got it! Thanks for the explanation. Going to merge this now, and create a new release!

PR #47, opened 2026-05-05
feat: rename `allow` config option to `auth`

tomaspozo:
## Summary

- Renames the `allow` config option to `auth` across `withSupabase`, `createSupabaseContext`, `verifyAuth`, and `verifyCredentials` to match Supabase CLI terminology — `auth: 'user'` reads more naturally than `allow: 'user'`.
- The legacy `allow` key still works for backward compatibility; using it emits a one-time `console.warn` per process. When both `auth` and `allow` are provided, `auth` wins. `allow` will be removed in a future major release.
- Adds new exported types `AuthMode` and `AuthModeWithKey`. `Allow` and `AllowWithKey` are kept as deprecated type aliases.
- Migration is a find-and-replace from `allow:` → `auth:` — the values are identical.

## Notes

- All call-sites in source, tests, README, `skills/supabase-server/SKILL.md`, and `docs/*.md` updated to use `auth:`.
- Added migration callouts at the top of `README.md`, `docs/auth-modes.md`, and `SKILL.md`.
- New `core/utils/deprecation.ts` centralizes the `auth` / `allow` resolution and the one-shot warn.
- New `allow → auth deprecation` test blocks in `verify-credentials.test.ts` and `with-supabase.test.ts` cover: `allow` still works, warning fires once, no warning when `auth` is used, `auth` precedence when both provided, default `'user'` when neither provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,32 +2,29 @@
 
 ## [0.1.4](https://github.com/supabase/server/compare/server-v0.1.3...server-v0.1.4) (2026-04-01)
 
-
 ### Features
 
-* add `supabaseOptions` and refactor client creation to options objects ([#19](https://github.com/supabase/server/issues/19)) ([5a10099](https://github.com/supabase/server/commit/5a100995a1b6254f92768c82c74b1c754c29b3b2))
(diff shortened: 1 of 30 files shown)


tomaspozo:
Restructuring PRs and scope

PR #53, opened 2026-05-07
feat(env): add support for JWKS discovery endpoints

awaseem:

Feature: Support for JWKS urls

##
- I added a inline variable that stores the remote cache from Jose when called once, Thinking here was that a subsequent serverless runtimes make requests in the warm pool we don't need to call it over and over again to reduce latency and rely on the underlying lib for caching 
- Added some tests with the help of claude

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,12 +400,13 @@ export default {
 
 Automatically available in Supabase Edge Functions:
 
-| Variable                    | Format                                                        | Description                                                                  |
-| --------------------------- | ------------------------------------------------------------- | ---------------------------------------------------------------------------- |
-| `SUPABASE_URL`              | `https://<ref>.supabase.co`                                   | Your project URL                                                             |
-| `SUPABASE_PUBLISHABLE_KEYS` | `{"default":"sb_publishable_...","web":"sb_publishable_..."}` | Publishable API keys (named)                                                 |
-| `SUPABASE_SECRET_KEYS`      | `{"default":"sb_secret_...","web":"sb_secret_..."}`           | Secret API keys (named)                                                      |
-| `SUPABASE_JWKS`             | `{"keys":[...]}`, `[...]`, or `https://...`                   | JSON Web Key Set for JWT verification (inline JSON or remote `https://` URL) |
+| Variable                    | Format                                                        | Description                                               |
+| --------------------------- | ------------------------------------------------------------- | --------------------------------------------------------- |
+| `SUPABASE_URL`              | `https://<ref>.supabase.co`                                   | Your project URL                                          |
+| `SUPABASE_PUBLISHABLE_KEYS` | `{"default":"sb_publishable_...","web":"sb_publishable_..."}` | Publishable API keys (named)                              |
+| `SUPABASE_SECRET_KEYS`      | `{"default":"sb_secret_...","web":"sb_secret_..."}`           | Secret API keys (named)                                   |
+| `SUPABASE_JWKS`             | `{"keys":[...]}` or `[...]`                                   | Inline JSON Web Key Set for JWT verification              |
+| `SUPABASE_JWKS_URL`         | `https://...`                                                 | Remote JWKS endpoint (used when `SUPABASE_JWKS` is unset) |
 
 Also supported (for local dev, self-hosted, or other runtimes):
 
--- a/README.md
+++ b/README.md
@@ -400,12 +400,12 @@ export default {
 
 Automatically available in Supabase Edge Functions:
 
-| Variable                    | Format                                                        | Description                           |
-| --------------------------- | ------------------------------------------------------------- | ------------------------------------- |
-| `SUPABASE_URL`              | `https://<ref>.supabase.co`                                   | Your project URL                      |
-| `SUPABASE_PUBLISHABLE_KEYS` | `{"default":"sb_publishable_...","web":"sb_publishable_..."}` | Publishable API keys (named)          |
-| `SUPABASE_SECRET_KEYS`      | `{"default":"sb_secret_...","web":"sb_secret_..."}`           | Secret API keys (named)               |
-| `SUPABASE_JWKS`             | `{"keys":[...]}` or `[...]`                                   | JSON Web Key Set for JWT verification |
+| Variable                    | Format                                                        | Description                                                                  |
+| --------------------------- | ------------------------------------------------------------- | ---------------------------------------------------------------------------- |
+| `SUPABASE_URL`              | `https://<ref>.supabase.co`                                   | Your project URL                                                             |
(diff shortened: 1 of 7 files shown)


docs/environment-variables.md
```diff
@@ -77,14 +77,17 @@ When both singular and plural forms are set, the plural form takes priority.
...
 
 ```
 # Standard JWKS format
 SUPABASE_JWKS={"keys":[{"kty":"RSA","n":"...","e":"AQAB"}]}
 
 # Bare array (convenience)
 SUPABASE_JWKS=[{"kty":"RSA","n":"...","e":"AQAB"}]
+
+# Remote JWKS URL — keys are fetched on demand and cached in memory.
+SUPABASE_JWKS=https://<ref>.supabase.co/auth/v1/.well-known/jwks.json
```
tomaspozo: @awaseem what do you think about making this a specific URL env var, like SUPABASE_JWKS_URL.

We first:
- Look for SUPABASE_JWKS
- Then look for SUPABASE_JWKS_URL
- Then use SUPABASE_URL + /auth/v1/.well-known/jwks.json

Imo, agents are not very well when it comes to conts and vars with mixed data types/purposes.
SUPABASE_JWKS -> Json object that provides actual keys
SUPABASE_JWKS_URL -> String that provides a fallback url.

awaseem: Lets do it, will update PR tomorrow!

awaseem:
Moving to draft because change in direction

docs/environment-variables.md
```diff
@@ -176,7 +187,8 @@ interface SupabaseEnv {
   url: string
   publishableKeys: Record<string, string>
   secretKeys: Record<string, string>
-  jwks: JsonWebKeySet | null
+  // `URL` when SUPABASE_JWKS is a remote endpoint, `JsonWebKeySet` for inline keys
+  jwks: JsonWebKeySet | URL | null
```
kallebysantos, line 195: Not sure if we should keep this one as separated property?
The enum type seems fine also.

awaseem: its only when it gets assigned down, when parsing the values the function makes the decision based on the env key

README.md
```diff
@@ -400,12 +400,13 @@ export default {
...
+| Variable                    | Format                                                        | Description                                               |
+| --------------------------- | ------------------------------------------------------------- | --------------------------------------------------------- |
+| `SUPABASE_URL`              | `https://<ref>.supabase.co`                                   | Your project URL                                          |
+| `SUPABASE_PUBLISHABLE_KEYS` | `{"default":"sb_publishable_...","web":"sb_publishable_..."}` | Publishable API keys (named)                              |
+| `SUPABASE_SECRET_KEYS`      | `{"default":"sb_secret_...","web":"sb_secret_..."}`           | Secret API keys (named)                                   |
+| `SUPABASE_JWKS`             | `{"keys":[...]}` or `[...]`                                   | Inline JSON Web Key Set for JWT verification              |
+| `SUPABASE_JWKS_URL`         | `https://...`                                                 | Remote JWKS endpoint (used when `SUPABASE_JWKS` is unset) |
```
kallebysantos: This env is not exposed – and I don't plan to do it - on Platform and other Supabase edfn envs

docs/environment-variables.md
```diff
@@ -2,25 +2,26 @@
...
+| Variable                    | Format                             | Description                                                  | Available in                      |
+| --------------------------- | ---------------------------------- | ------------------------------------------------------------ | --------------------------------- |
+| `SUPABASE_URL`              | `https://<ref>.supabase.co`        | Your Supabase project URL                                    | All                               |
+| `SUPABASE_PUBLISHABLE_KEYS` | `{"default":"sb_publishable_..."}` | Named publishable keys as JSON object                        | All                               |
+| `SUPABASE_SECRET_KEYS`      | `{"default":"sb_secret_..."}`      | Named secret keys as JSON object                             | All                               |
+| `SUPABASE_JWKS`             | `{"keys":[...]}` or `[...]`        | Inline JSON Web Key Set for JWT verification                 | All                               |
+| `SUPABASE_JWKS_URL`         | `https://...`                      | Remote JWKS endpoint (alternative to inline `SUPABASE_JWKS`) | All                               |
```
kallebysantos: This won't be available on Supabase envs, and should be only referenced in the "Non-Supabase" table

kallebysantos (changes requested):
Looks good to me, 
We just need to make sure to this env is not referenced as default exposed.

docs/environment-variables.md
awaseem: Good point — went back and forth on this. Kept the union (`JsonWebKeySet | URL | null`) because resolution is mutually exclusive: only one of `SUPABASE_JWKS` / `SUPABASE_JWKS_URL` ever wins. After `resolveEnv` there's exactly one JWKS source, and `verify-credentials.ts` only needs to discriminate once via `instanceof URL`. Splitting into two fields on `SupabaseEnv` would push that branching onto every consumer and turn "which one is authoritative?" into an open question on the type, instead of a closed question at the resolution layer. Happy to revisit if you'd rather mirror the env-var split on the type — let me know.

src/core/resolve-env.ts
```diff
@@ -77,12 +77,52 @@ function parseJwks(raw: string | undefined): JsonWebKeySet | null {
...
+ * Parses a JWKS endpoint URL. Only `https://` is accepted: a MITM on the
+ * JWKS fetch could swap in an attacker-controlled key and forge JWTs that
+ * pass verification. Returns `null` for missing, non-https, or malformed input.
+ *
+ * @internal
+ */
+function parseJwksUrl(raw: string | undefined): URL | null {
+  if (!raw) return null
+  const trimmed = raw.trim()
+  if (!trimmed.startsWith('https://')) return null
```
tomaspozo: What about local dev with the cli? Since we offer local dev with the cli, I think we should also accept those. Clauded this, wdyt?

```ts
function parseJwksUrl(raw: string | undefined): URL | null {                                                         
  if (!raw) return null                                                                                              
  const trimmed = raw.trim()
  try {                                                                                                              
    const url = new URL(trimmed)
    if (url.protocol === 'https:') return url                                                                        
    if (url.protocol === 'http:' && isLoopbackHost(url.hostname)) return url
    return null                                                                                                      
  } catch {
    return null                                                                                                      
  }                      
}

function isLoopbackHost(hostname: string): boolean {                                                                 
  // URL parsing strips IPv6 brackets, so we get "::1" not "[::1]"
  if (hostname === 'localhost' || hostname.endsWith('.localhost')) return true                                       
  if (hostname === '::1') return true                                                                                
  if (/^127\.\d{1,3}\.\d{1,3}\.\d{1,3}$/.test(hostname)) return true                                                 
  return false                                                                                                       
}    
```

tomaspozo (commented):
A small one proposing to support local dev urls

kallebysantos (approved):

PR #54, opened 2026-05-07
feat: add user-scoped Supabase auth

Albert-Gao:
## Summary

- add `verifyUserAuth` as a core helper for Supabase user JWT verification
- default user-token audience validation to `authenticated` and support optional user ID matching
- add Hono `withSupabaseUserAuth` middleware that sets a user-scoped Supabase context without creating an admin client

## Why

`withSupabase` and `createSupabaseContext` always create `supabaseAdmin`, which requires `SUPABASE_SECRET_KEY`. User-facing runtime APIs often only need a verified Supabase user token and an RLS-scoped client, so this adds a narrower path that does not require secret keys in those runtimes.

## Validation

- `pnpm exec prettier --check src/adapters/hono/index.ts src/adapters/hono/middleware.test.ts src/adapters/hono/middleware.ts src/core/index.ts src/core/verify-user-auth.test.ts src/core/verify-user-auth.ts src/index.ts src/types.ts`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `pnpm build`

### DIFF
--- a/src/adapters/hono/index.ts
+++ b/src/adapters/hono/index.ts
@@ -4,4 +4,4 @@
  * @packageDocumentation
  */
 
-export { withSupabase } from './middleware.js'
+export { withSupabase, withSupabaseUserAuth } from './middleware.js'
--- a/src/adapters/hono/middleware.test.ts
+++ b/src/adapters/hono/middleware.test.ts
@@ -1,10 +1,17 @@
+import { exportJWK, generateKeyPair, SignJWT } from 'jose'
 import { Hono } from 'hono'
-import { describe, expect, it } from 'vitest'
+import { beforeAll, describe, expect, it } from 'vitest'
 
-import type { SupabaseContext } from '../../types.js'
-import { withSupabase } from './middleware.js'
+import type {
+  JsonWebKeySet,
+  SupabaseContext,
+  SupabaseEnv,
+  SupabaseUserContext,
+} from '../../types.js'
+import { withSupabase, withSupabaseUserAuth } from './middleware.js'
 
 type Env = { Variables: { supabaseContext: SupabaseContext } }
+type UserEnv = { Variables: { supabaseUserContext: SupabaseUserContext } }
 
 describe('hono supabase middleware', () => {
   const env = {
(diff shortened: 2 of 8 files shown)


kallebysantos:
Hey @Albert-Gao,
May you share a use case where you need this kind of `ID` verification using the currently `withSupabase()` flow?
I got the idea, but I'm not really sure if creating a new hook function just for it is the correct path.

Albert-Gao:
It is a use case that i am creating dedicated resources for each user with his own VM, in this VM, the server needs to auth the user, so I need to make sure that not only it is a valid Supabase token, but it is a token minted for that particular user.

If the api can provide us a way to validate the claims with our own code, that would eliminate this PR with a more flexible model. It also means on my end, I just need a smaller wrapper.

@kallebysantos

kallebysantos:
Yes sure, but are you expecting to get this user Id from request pathname or somewhere else?
I still confused with the `userId` been supplied as [hardcoded value](https://github.com/Albert-Gao/server/blob/cbfd0bd5186e714f40123f255ec1209fbde04904/src/adapters/hono/middleware.test.ts#L151).

kallebysantos:
> If the api can provide us a way to validate the claims with our own code....I just need a smaller wrapper.

Yh its a valid request, I did already pushed a discussion with team to see if we found a nice approach for it

tomaspozo:
Hey @Albert-Gao!

Thanks for the contribution and the use case write up. After talking it through with the team, we'd rather not land this in core, but I think you can solve your case with a small wrapper.

The `userId` check is essentially a one-liner in user code, and in real apps the expected ID has to come from the request (path param, session, etc.), which is awkward to express as middleware config. On the "no `supabaseAdmin` so no secret key" part, we'll take that into further consideration; if you'd like, please open an issue so we can keep you updated there.

Here is an idea for Hono using the existing primitives:

```ts
// auth.ts
import type { MiddlewareHandler } from 'hono'
import { HTTPException } from 'hono/http-exception'
import { withSupabase } from '@supabase/server/adapters/hono'
import type { SupabaseContext } from '@supabase/server'

export function withScopedUser(
  resolveExpectedUserId: (c: Parameters<MiddlewareHandler>[0]) => string,
): MiddlewareHandler<{ Variables: { supabaseContext: SupabaseContext } }> {
  const base = withSupabase({ auth: 'user' })

  return async (c, next) => {
    await base(c, async () => {
      const expected = resolveExpectedUserId(c)
      const actual = c.var.supabaseContext.userClaims?.id

      if (actual !== expected) {
        throw new HTTPException(401, { message: 'Token does not match user' })
      }

      await next()
    })
  }
}
```

```ts
// app.ts
const app = new Hono()

app.use(
  '/users/:userId/*',
  withScopedUser((c) => c.req.param('userId')),
)
```

We are working on adding more documentation and examples on how to extend `withSupabase` for custom use cases like this one.

Closing this PR for now, thanks again for taking the time to put it together!

Albert-Gao:
@kallebysantos  ye, it is a hard coded value since for each VM, the user id will be injected as env var, so in the end, it does not matter where you are getting that string, you can just pass the final value to the middleware, we basically just need to check if the user id from the token is the same as the user id in the env var

@tomaspozo hey, thanks for the example, will do that for now! was basically trying to avoid it on my end xD

PR #55, opened 2026-05-08
feat: add NestJS adapter

bogdantarasenko:
## Summary

Adds a community NestJS adapter under `@supabase/server/adapters/nestjs`, alongside the existing Hono and H3 adapters.

- `withSupabase(opts)` returns a class guard usable with `@UseGuards()` or `app.useGlobalGuards(new (withSupabase(...))())`. Works on both Express and Fastify platforms; HTTP/2 pseudo-headers and non-HTTP execution contexts (rpc/ws) are handled.
- `@SupabaseCtx(key?, ...pipes)` is a param decorator that returns the full `SupabaseContext` or a single field, with NestJS pipes applied to the extracted value.
- Failed auth surfaces as `HttpException` with `{ message, code }`; the underlying `AuthError` is exposed on `cause` so consumers can build their own exception filters.

`@nestjs/common` is added as an **optional** peer dep (`^10.0.0 || ^11.0.0`), so users who don't import the adapter aren't forced to install Nest.

## What's in the diff

- `src/adapters/nestjs/{middleware,decorator,index}.ts` — adapter source
- `src/adapters/nestjs/{middleware,integration}.test.ts` — unit tests for the guard/decorator plus integration tests that boot a real Nest app on Express *and* Fastify (covers `@UseGuards`, `useGlobalGuards`, and `@SupabaseCtx` with pipes)
- `vitest.config.ts` — uses `unplugin-swc` so the integration tests get `emitDecoratorMetadata`, which Nest's DI requires (esbuild, vitest's default transform, doesn't emit it)
- `tsconfig.json` — `experimentalDecorators` + `emitDecoratorMetadata` for the same reason
- `tsdown.config.ts`, `package.json` exports, `jsr.json` — wire the new entry point
- `README.md`, `src/adapters/README.md`, new `docs/adapters/nestjs.md` — documentation

## Test plan

- [x] `pnpm test` — unit + integration tests pass on Express and Fastify
- [x] `pnpm build` — `tsdown` emits `dist/adapters/nestjs/index.{mjs,cjs,d.mts,d.cts}`
- [ ] Reviewer: confirm the optional-peer-dep approach (`peerDependenciesMeta`) is acceptable so non-Nest users aren't affected
- [ ] Reviewer: sanity-check the `vitest.config.ts` + `unplugin-swc` test setup; it's only needed because Nest's DI requires decorator metadata that esbuild doesn't emit


tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "module": "ESNext",
     "moduleResolution": "bundler",
     "strict": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true
```
mandarini: These two settings should be moved in a new scoped file:

```text
src/adapters/nestjs/tsconfig.json 
```

where you can write somethign like 

```json
  {
    "extends": "../../../tsconfig.json",
    "compilerOptions": {
      "experimentalDecorators": true,
      "emitDecoratorMetadata": true
    }
  }
```

so that these two settings are not enforced on the whole repo.

vitest.config.ts
```diff
@@ -0,0 +1,16 @@
+import swc from 'unplugin-swc'
+import { defineConfig } from 'vitest/config'
+
+// SWC handles legacy decorators with metadata for the NestJS integration test.
+// The default vitest transform (esbuild) doesn't emit decorator metadata, which
+// NestJS needs to wire up controllers, guards, and DI.
```
mandarini: Ack, we can remove this comment

vitest.config.ts
```diff
@@ -0,0 +1,16 @@
...
+export default defineConfig({
+  plugins: [
+    swc.vite({
+      jsc: {
+        parser: { syntax: 'typescript', decorators: true },
+        transform: { decoratorMetadata: true, legacyDecorator: true },
+      },
+    }),
+  ],
```
mandarini: I think I would prefer here also to scope the swc setting only where it's needed (nestjs), so I think it would be better if we used the projects syntax:

```ts
  export default defineConfig({
    test: {                                                                                                                                     
      projects: [ ... ],
    },
  })
```

so project 1 that would target all the tests would be like:

```ts
  {
    test: {
      name: 'unit',
      include: ['src/**/*.test.ts'],
      exclude: ['src/adapters/nestjs/**'],
    },
  }, 
```

and project 2 for nest would be like

```ts
  {                                                                                                                                             
    plugins: [    
      swc.vite({
        jsc: {  
          parser: { syntax: 'typescript', decorators: true },
          transform: { decoratorMetadata: true, legacyDecorator: true },                                                                        
        },                                                              
      }),                                                                                                                                       
    ],            
    test: {
      name: 'nestjs',
      include: ['src/adapters/nestjs/**/*.test.ts'],                                                                                            
    },                                              
  },    
```

src/adapters/nestjs/middleware.ts
```diff
@@ -0,0 +1,119 @@
...
+  // bundled output contains no decorator tokens. Several JS runtimes (incl.
+  // Node's CJS loader) refuse to parse decorator syntax at runtime, even
+  // though the metadata semantics are identical.
+  class SupabaseAuthGuard implements CanActivate {
+    async canActivate(executionContext: ExecutionContext): Promise<boolean> {
+      // HTTP-only: this guard reads HTTP headers via `switchToHttp()`. In RPC
+      // or WebSocket contexts the request shape differs (no `headers`), so
+      // skip rather than crash. Users on those transports should authenticate
+      // via the appropriate context-specific mechanism.
+      if (executionContext.getType() !== 'http') return true
```
mandarini: I have a small concern about the fail-open. Returning true for non-HTTP contexts means the guard silently allows every message on a WS gateway or RPC handler, so if someone wires it onto one of those expecting auth, they'd get a no-op with no runtime signal. I see the comment explaining the HTTP-only scope, which is totally fair, the guard genuinely can't read non-HTTP requests. But the skip behavior is what worries me, not the scoping. Two optiosn that I think are safer: 
1. Throw on non-HTTP contexts so misuse fails loudly on the first request rather than silently passing through. Same "we don't authenticate transports we can't read" intent, but enforced instead of advisory.
2. Return false to deny by default.                                                                                                           

If the current return true is the deliberate choice, I'd at least want the JSDoc on withSupabase to call out the HTTP-only constraint prominently. Right now a reader scanning the API could reasonably assume the guard authenticates whatever Nest hands it.

src/adapters/nestjs/middleware.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      const req = executionContext.switchToHttp().getRequest<NestRequestLike>()
+
+      // Skip if a previous guard already set the context. Enables stacking
+      // `@UseGuards(withSupabase({ auth: 'user' }))` at the controller level
+      // with a different auth mode at the handler level — the first one wins.
+      if (req.supabaseContext) return true
```
mandarini: This makes me a bit nervous. Nest's guard order is fixed (global → controller → handler), so "first wins" here always means "outermost wins". Handler-level guards can never tighten what a global guard set. This, I think, could grant a user token access to a route the handler explicitly scoped to secret.

mandarini (changes requested):
Hi @bogdantarasenko ! Thank you very much for this PR and for contributing to Supabase! :D I left some change requests, can you please take a look?

bogdantarasenko:
Thanks for the review @mandarini — all five points addressed in 8552a11.

**1. Scoped decorator tsconfig** — `experimentalDecorators` and `emitDecoratorMetadata` removed from the root `tsconfig.json`; new `src/adapters/nestjs/tsconfig.json` extends root and enables them only for the adapter. The root config now excludes `src/adapters/nestjs`, and `pnpm typecheck` runs both projects (`tsc --noEmit && tsc --noEmit -p src/adapters/nestjs`). The scoped include also picks up `../../env.d.ts` since the adapter transitively imports `resolve-env.ts` which uses the `Deno` global.

**2. vitest `projects` for swc** — `vitest.config.ts` now defines a `unit` project (excludes `src/adapters/nestjs/**`) and a `nestjs` project that loads `unplugin-swc`. Removed the top-of-file comment. Output now reports each project's tests under its label.

**3. Non-HTTP contexts now throw** — went with option 1 (fail loudly). The guard throws `HttpException(500, { code: 'unsupported_context' })` referencing the actual context type, instead of returning true. JSDoc on `withSupabase` also calls out the HTTP-only constraint prominently at the top. Test updated to assert the throw on both `rpc` and `ws`.

**4. \"Outermost wins\" guard escalation fixed** — agreed, that was a real escalation hazard given Nest's fixed global → controller → handler order. Removed the `if (req.supabaseContext) return true` skip; each guard now always re-evaluates against its own configured auth mode, so a handler-level guard can tighten what a global one set. The innermost guard's context wins. The old \"skips if context already set\" test was replaced with two tests: one proves a stricter handler guard rejects when a permissive outer guard had populated context, the other proves successful inner auth overwrites the preset.

Also cleaned up a stale unused `CanActivate` import in `integration.test.ts` while in the area. All 154 tests pass; `pnpm typecheck`, `pnpm lint`, and `pnpm build` are clean.

mandarini (changes requested):
Hi @bogdantarasenko, thank you so much for the thorough follow-up on this! You addressed every point really carefully, and the two new tests around guard re-evaluation (inner-rejects + inner-overwrites) make the intended semantics very clear. The throw-on-non-HTTP fix is exactly what I was hoping for, and the scoped `tsconfig.json` + vitest `projects` split is clean.

One last thing before we can merge: `docs/adapters/nestjs.md` still has the old "Skip behavior" section (lines 118-122) describing the skip-if-context-already-set behavior, which no longer matches the code after this round. The JSDoc on `withSupabase` already has the correct wording ("Always runs, even if a previous guard already set the context..."); could you rewrite that doc section to match? The closing advice about preferring per-route `@UseGuards(...)` without a global guard is still good and worth keeping.

Thanks again for sticking with this through two rounds of review, it is contributions like yours that make Supabase what it is.

bogdantarasenko:
Done — pushed in 3c54f65. Rewrote the section (now `## Multiple guards`) to match the JSDoc: the guard always runs, NestJS's global → controller → handler order means the innermost guard wins (either rejecting or overwriting), and kept the closing advice about per-route `@UseGuards(...)` without a global guard. Thanks again for the careful reviews!

mandarini (approved):

mandarini:
Thank you for the fixes @bogdantarasenko !

tomaspozo (approved):

PR #66, opened 2026-05-15
feat: add Express 5 adapter

bogdantarasenko:
## Summary

Adds a community Express 5 adapter under `@supabase/server/adapters/express`, alongside the existing Hono, H3, and Elysia adapters.

- `withSupabase(config)` returns a standard Express middleware. Resolved context lives on `res.locals.supabaseContext` via `Express.Locals` declaration merging, so downstream handlers and middleware get full type inference.
- `requireAuth(...modes)` is a per-route guard that asserts the context is set and (optionally) that `authMode` matches one of the listed modes. Useful when one app mixes routes with different auth requirements.
- `withSupabaseRoute(config, handler)` is a per-route wrapper for cases where global middleware doesn't fit — same context shape, async errors are forwarded to `next()`.
- Auth errors flow through a configurable `onError` option. Default is `next(error)` (Express-idiomatic); pass a function to send a custom response instead.
- The Express `req` → Fetch `Request` bridge preserves headers (incl. multi-value `Set-Cookie` / `Cookie`), composes absolute URLs under `app.set('trust proxy', true)`, and forwards bodies for non-GET/HEAD methods.

`express` is added as an **optional** peer dep (`^5.0.0`), so users who don't import the adapter aren't forced to install Express. No new runtime dependencies.

## What's in the diff

- `src/adapters/express/{middleware,require-auth,with-supabase-route,to-fetch-request,index}.ts` — adapter source
- `src/adapters/express/*.test.ts` — unit tests covering every `AuthMode` (`user` / `publishable` / `secret` / `none`), the array form, invalid-JWT / missing-credential failures, the `res.locals.supabaseContext` short-circuit path, default and custom `onError`, the `requireAuth()` guard (present / missing context / mode mismatch), the `withSupabaseRoute()` wrapper (success, auth failure, async-thrown errors), and the `req` → `Request` bridge (headers, body, trust-proxy URL composition)
- `tsdown.config.ts`, `package.json` exports + `peerDependencies` + `peerDependenciesMeta` + `devDependencies`, `jsr.json` — wire the new entry point
- `README.md`, `src/adapters/README.md`, new `docs/adapters/express.md` — documentation

## Test plan

- [x] `pnpm install` resolves the new peer/dev dependencies
- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm test` — 190/190 pass, including all Express adapter test files
- [x] `pnpm build` produces `dist/adapters/express/index.{mjs,cjs,d.mts,d.cts}`
- [ ] Reviewer: confirm the optional-peer-dep approach (`peerDependenciesMeta`) is consistent with the other adapters
- [ ] Reviewer: smoke-test on an Express 5 app with `app.set('trust proxy', true)` behind a reverse proxy to confirm the bridged URL uses `X-Forwarded-*`

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,11 +266,12 @@ Adapters wrap `withSupabase` for a specific framework's middleware contract. The
 
 > **Adapters are a community-driven initiative.** They're developed, maintained, and evolved by contributors — including responding to upstream framework changes. See [`src/adapters/README.md`](src/adapters/README.md) for the contribution requirements (tests, types, docs, build wiring) if you'd like to add or help maintain one.
 
-| Framework | Import                             | Framework version | Docs                                               |
-| --------- | ---------------------------------- | ----------------- | -------------------------------------------------- |
-| Hono      | `@supabase/server/adapters/hono`   | `^4.0.0`          | [docs/adapters/hono.md](docs/adapters/hono.md)     |
-| H3 / Nuxt | `@supabase/server/adapters/h3`     | `^2.0.0`          | [docs/adapters/h3.md](docs/adapters/h3.md)         |
-| Elysia    | `@supabase/server/adapters/elysia` | `^1.4.0`          | [docs/adapters/elysia.md](docs/adapters/elysia.md) |
+| Framework | Import                              | Framework version | Docs                                                 |
+| --------- | ----------------------------------- | ----------------- | ---------------------------------------------------- |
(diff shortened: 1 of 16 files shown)


kallebysantos:
Please remove this `.chief/` folder, I believe it was accidentally added

PR #78, opened 2026-06-12
chore(types): export common set of types for supabase server

awaseem:

- Export supabase js common libs so you don't need to install locally to resolve


kallebysantos (approved):

mandarini:
Nice!! I agree with @kallebysantos on the `/peer/supabase-js` subpath, that's the right shape. I do have a few small notes before merging 😅 :

On the PR description framing: `supabase-js` is a peer dep, so users install it either way. What's nice here isn't skipping the install, it's having one import instead of two. The jsdoc in `src/peer/supabase-js/index.ts` says this well, I think we should just align the PR body with it.

Another note is that this is now a curated surface to be maintained. New types added to supabase-js won't auto show up. I think it's fine (shows the intention), but let's be explicit so we don't ask "can you add type X" every time someone asks.

I like that we're making an alias of `AuthError` to `SupabaseAuthError`. It could be worth a note (just one line) in `docs/api-reference.md` so users know why it's renamed.

Otherwise, this is great! :D

kallebysantos:
Thanks @mandarini 

> It could be worth a note (just one line) in docs/api-reference.md so users know why it's renamed.

I've reverted it. Better avoid hidden behaviors like renaming, etc. But included the alias in the example so users know they must avoid collision

PR #83, opened 2026-06-17
feat: add support for `HS256` JWKs

kallebysantos:

Feature, Bug fix

Current Supabase's Signing Keys `/auth/v1/.well-known/jwks.json` supports `HS256` algorithm but Server SDK don't

## What is the new behavior?

This PR add support for Symmetric Shared Secret validation 

## Additional context

Towards FUNC-702

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -100,6 +100,7 @@ export interface SupabaseEnv {
    * Each env var is authoritative when set: a malformed value resolves to
    * `null` rather than falling through to the other variable.
    */
+  // TODO:(kallebysantso) This should map to jose `JSONWebKeySet`
   jwks: JsonWebKeySet | URL | null
 }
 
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -100,6 +100,7 @@ export interface SupabaseEnv {
    * Each env var is authoritative when set: a malformed value resolves to
    * `null` rather than falling through to the other variable.
    */
+  // TODO:(kallebysantso) This should map to jose `JSONWebKeySet`
```
tomaspozo: @kallebysantos lets create an internal issue on linear for this, so we don't leave todos in code

tomaspozo (approved):

src/types.ts
kallebysantos: created FUNC-705

PR #91, opened 2026-07-03
refactor: replace custom JsonWebKeySet with JSONWebKeySet

mandarini:
The custom JsonWebKeySet type used the browser's built-in JsonWebKey[] which lacks standard JWK fields like kid. Jose's JSONWebKeySet uses JWK[] which is a proper superset. No public API change, the type was internal.


kallebysantos (approved):
Thanks!! 💚

PR #99, opened 2026-07-08
test: add E2E tests for all four adapters against a local Supabase stack

mandarini:
Adds an e2e vitest project covering what the mocked unit tests cannot: real GoTrue-issued JWTs verified against the live JWKS endpoint, real Supabase client operations via supabaseAdmin, resolveEnv() reading process.env, and imports from dist/ so packaging regressions fail here.

One scenario set (auth + data access + isolation) runs over real HTTP against minimal Hono, H3, Elysia, and NestJS apps. Elysia runs behind a node:http server (srvx) so CI needs no Bun. A separate E2E workflow starts the local stack with the Supabase CLI, builds, and runs the suite.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,39 @@
+# E2E tests
+
+End-to-end coverage for `@supabase/server`: real GoTrue-issued JWTs, real JWKS
+validation over HTTP, real Supabase client operations — across all four
+adapters (Hono, H3, Elysia, NestJS).
+
+Unlike the unit/integration tests (mocked env, `jwks: null`), this suite:
+
+- imports the library from `dist/`, not `src/`, so packaging regressions fail here
+- reads config from `process.env` via `resolveEnv()` — no mocked env objects
+- verifies JWTs against the local stack's live JWKS endpoint
+- runs each adapter app on a real HTTP server and asserts over `fetch`
+
+## Running locally
+
+```sh
+pnpm build                  # e2e imports from dist/
+cd e2e && supabase start    # local stack (Docker) on ports 5433x
+cd .. && pnpm gen:env       # writes e2e/.env from `supabase status`
+pnpm test:e2e
+```
+
+Run a single adapter with `pnpm test:e2e h3`.
+
+## Layout
+
+- `supabase/` — local stack config + `notes` table migration
+- `apps/<adapter>/app.ts` — minimal app per adapter, identical route surface:
+  `GET /health` (public), `GET /me` (user), `GET /me-optional` (user or none),
+  `GET|POST /notes` (user, admin client scoped by `userClaims.id`)
+- `scenarios.ts` — the single scenario set run against every adapter
+- `setup/global-setup.ts` — checks the stack is up, signs in two test users,
+  provides their tokens to the tests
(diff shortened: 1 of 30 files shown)


grdsdev (approved):
very good! 👏

mandarini:
Thanks! I also want @kallebysantos, @tomaspozo and @awaseem to take a look!

e2e/apps/notes.ts
```diff
@@ -0,0 +1,54 @@
...
+): Promise<NoteRow[]> {
+  const { data, error } = await supabase
+    .from('notes')
+    .select(COLUMNS)
+    .order('created_at', { ascending: true })
+  if (error) throw new Error(`list own notes failed: ${error.message}`)
+  return data as unknown as NoteRow[]
+}
+
+/** Lists only the calling user's notes. */
```
kallebysantos, line 57: Maybe it would be better to list all? This way, it ensures that Admin is not user-scoped?

e2e/setup/token.ts
```diff
@@ -0,0 +1,59 @@
...
+    const { error: signUpError } = await supabase.auth.signUp({
+      email,
+      password,
+    })
+    if (signUpError) {
+      throw new Error(
+        `could not create test user ${email}: ${signUpError.message}`,
+      )
+    }
+    ;({ data, error } = await supabase.auth.signInWithPassword({
```
kallebysantos: Does this `;` mean something, or is it just a typo? And is the `()` part of the value assignment syntax?

e2e/README.md
```diff
@@ -0,0 +1,45 @@
+# E2E tests
+
+End-to-end coverage for `@supabase/server`: real GoTrue-issued JWTs, real JWKS
+validation over HTTP, real Supabase client operations — across all four
+adapters (Hono, H3, Elysia, NestJS).
```
kallebysantos: Are we focusing on testing only the adapters? How about spinning up an edge function? This will ensure the full-stack e2e with Supabase CLI

kallebysantos (approved):
Looks good, just a few comments

awaseem (approved):
Much needed! Thank you <3

e2e/apps/notes.ts
mandarini: Good catch, added `GET /all-notes` in ed5bb26: it reads through `supabaseAdmin` with no filter, and the scenario asserts user2's request sees user1's rows. That directly proves the admin client isn't scoped to the caller. Kept `GET /notes` filtered as the realistic app-layer authorization example.

e2e/setup/token.ts
mandarini: Intentional but ugly 😄, the parens are required when destructuring into existing variables, and the leading `;` prevents ASI from gluing the `(` onto the previous line (no-semicolon codebase). Refactored it away in ed5bb26.

e2e/README.md
mandarini: Good point, the core `withSupabase(config, handler)` wrapper (what an Edge Function deploys) wasn't covered. ed5bb26 adds a fifth app running it against the full scenario set. The real Deno runtime via `supabase functions serve`: i added a new issue

PR #100, opened 2026-07-08
fix: add hono adapter Database type override

### LINKED ISSUE #101: hono adapter type override improvment
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The `withSupabase` hono adapter could use some slightly better type ergonomics. As a core feature, Hono supports chaining which compounds the type on the app as a declarative flow. It's not required, but is very useful for things like [RPC](https://hono.dev/docs/guides/best-practices#if-you-want-to-use-rpc-features) and using the [zod-openapi](https://hono.dev/examples/zod-openapi) controller.

If we had a way to override the Database types when invoking `withSupabase`, the types would be retained through the chained declaration.

## Additional context

Maybe "bug" isn't the right context for this, but this feels more like a non-breaking patch bump than a minor version bump.

elGatoMantocko:

Fix hono adapter type definitions.

There is no way to override the DB context returned by the `withSupabase` middleware without doing so in the Hono app itself.

## What is the new behavior?

Make the hono `withSupabase` adapter more type-aware. Allow Database type overrides in the hono middleware.

## Additional context

For growing projects, managing routes with hono's [route helpers](https://hono.dev/docs/helpers/route#working-with-sub-applications) via sub-applications generally becomes best practice when scalaing a hono api. In those scenarios, it's common to see many hono "apps" or routes defined across a stack. This change allows users to just use the hono middleware without needing to pre-declare any context types.

### DIFF
--- a/docs/typescript-generics.md
+++ b/docs/typescript-generics.md
@@ -101,6 +101,26 @@ app.get('/todos', async (c) => {
 })
 ```
 
+Alternatively, skip the `Env` type and thread `Database` through the `withSupabase<Database>()` generic when using chained declarations. Hono infers the `supabaseContext` variable type from the middleware, so `c.var.supabaseContext.supabase` is still a fully typed `SupabaseClient<Database>`. This plays a little nicer with hono [sub-application routes](https://hono.dev/docs/helpers/route#working-with-sub-applications).
+
+```ts
+import { Hono } from 'hono'
+import { withSupabase } from '@supabase/server/adapters/hono'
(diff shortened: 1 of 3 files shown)


elGatoMantocko:
fixes #101

mandarini (approved):

PR #116, opened 2026-08-21
fix: migrate `withOAuthProtectedResource` to `defineMiddleware`

raulb:

Fixes [AI-995](https://linear.app/supabase/issue/AI-995/migrate-withoauthprotectedresource-to-definemiddleware) to unblock [AI-1047](https://linear.app/supabase/issue/AI-1047/make-withoauthprotectedresource-work-off-edge-functions)

`withOAuthProtectedResource` has no config parameter at the moment. To do AI-1047, it needs to add one so it can run on other backends. Bolting that onto a plain closure means hand-rolling config-optional/required overloads from scratch.

Also, any path other than `/{fn}` or `/{fn}/oauth-protected-resource` gets a blanket `404` — an accidental side effect of the closure, never a deliberate contract. See ([@mandarini's comment](https://linear.app/supabase/issue/AI-995/migrate-withoauthprotectedresource-to-definemiddleware#comment-cef00601)).

## What is the new behavior?

Also fixes `withSupabase` composition. `defineMiddleware` hands the inner handler a `ctx` object rather than the raw platform arg, so `withSupabase` no longer re-seeds over it: it now detects an existing context (via `isContext`, exported in `@supabase/middleware@0.3.1`) and spreads it instead. 

This keeps `withOAuthProtectedResource(withSupabase({...}, handler))` working, upstream ctx keys survive, and platform env stays intact (which would otherwise silently break `getEnv` on Workers).

Verified end-to-end against the notes-app example on the local Deno edge runtime: auth + tools suites pass, and a real MCP client OAuth flow (DCR + PKCE + tool calls) works.

### DIFF
--- a/e2e/supabase/functions/server-e2e/deno.json
+++ b/e2e/supabase/functions/server-e2e/deno.json
@@ -5,7 +5,7 @@
     "@supabase/server/middleware/postgres-admin": "../_vendor/package/dist/middleware/postgres-admin/index.mjs",
     "@supabase/supabase-js": "npm:@supabase/supabase-js@2",
     "@supabase/supabase-js/cors": "npm:@supabase/supabase-js@2/cors",
-    "@supabase/middleware": "npm:@supabase/middleware@0.3.0",
+    "@supabase/middleware": "npm:@supabase/middleware@0.3.1",
     "jose": "npm:jose@6",
     "pg": "npm:pg@8"
   }
--- a/package.json
+++ b/package.json
@@ -243,7 +243,7 @@
     "vitest": "^4.1.0"
   },
   "dependencies": {
-    "@supabase/middleware": "^0.3.0",
+    "@supabase/middleware": "^0.3.1",
     "jose": "^6.2.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12,8 +12,8 @@ importers:
   .:
     dependencies:
       '@supabase/middleware':
-        specifier: ^0.3.0
-        version: 0.3.0
+        specifier: ^0.3.1
+        version: 0.3.1(typescript@5.9.3)
       jose:
         specifier: ^6.2.0
         version: 6.2.0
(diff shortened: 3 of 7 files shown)


mandarini (changes requested):
Hey @raulb ! The conversion itself looks right, the `defineMiddleware` shape, the config param, and the 404→passthrough change are all good, and I'm glad the passthrough decision got made deliberately with tests. Two things need to change before merge though, and two smaller asks. Both real issues share one root cause, so I'll start there.

**Root cause: `withSupabase` rebuilds the context from scratch.** In `src/with-supabase.ts` (~line 181) it does `{ ...seedContext(platformArg), ... }` unconditionally. When it's nested under your new entry, `platformArg` is no longer the host's env — it's the marked ctx object your middleware built. Two consequences:

**1. The docstring's own example doesn't work.** `withOAuthProtectedResource(withSupabase({...}, handler))` — the handler can never read `ctx.oauthProtectedResource`, because `withSupabase` drops all upstream ctx keys when it rebuilds. So the contribution this PR adds is dead in the exact composition the docstring advertises. Please either:

- make the composition real: have `withSupabase` detect an upstream ctx and spread it into the context it builds instead of reseeding (see fix options below), **or**
- if that's out of scope here, change the docstring to an example that actually works today, and say explicitly that `ctx.oauthProtectedResource` doesn't survive `withSupabase` nesting yet.

**2. The Workers env clobber you flagged in the PR body.** Same root cause: `seedContext()` stashes whatever object it's given as the module-scoped platform env, so `seedContext(ctx)` overwrites the real Workers bindings that your entry correctly captured. You proposed the right fix yourself — guard on "is this already a context?".

**The fix** (one change covers both): `@supabase/middleware@0.3.1`, just released, exports `isContext` for exactly this. Bump the dependency, then in `with-supabase.ts` stop reseeding when the second argument is already a context:

```ts
const ctx = isContext(platformArg) ? platformArg : seedContext(platformArg)
```

and build `withSupabase`'s context by spreading `ctx` (so upstream keys like `ctx.oauthProtectedResource` survive) instead of starting from scratch. That un-clobbers the Workers bindings and makes your docstring example real in one move. A small test for the nested composition would be great too, but no blocker.

**Smaller asks:**

- Retitle `refactor:` → `fix:`. this PR changes runtime behavior (the 404→passthrough), and release-please won't surface a `refactor:` in the changelog.
- Can you please rebase with latest `main`? I'd like #116 to go into 1.5.0 too, so the API ships correct the first time.

Everything else looked good! the test coverage on the passthrough change in particular. Please lmk if I've misread the composition flow anywhere!

mandarini (approved):

PR #117, opened 2026-08-24
feat: `withOAuthProtectedResource` url config

raulb:

Feature (plus some bug fixes to the existing URL parsing).

`withOAuthProtectedResource` hardcodes the Supabase Edge Functions proxy shape into its URL derivation: it parses `X-Forwarded-Host/-Proto/-Port` and reconstructs the `/functions/v1` prefix the proxy strips. That's correct on Edge Functions but wrong anywhere else, off-platform the handler receives the request at its real external URL, and the Supabase project's origin is unrelated to the app's. It also takes no configuration at all, so there's no way to correct it.

Fixes https://linear.app/supabase/issue/AI-1047/make-withoauthprotectedresource-work-off-edge-functions

## What is the new behavior?

Two options, both accepting a fixed string or a function of the request, passed through the middleware's own config channel:

- **`resourceServer`**:  the resource identifier to advertise (this endpoint's externally-visible URL).
- **`authorizationServer`**: the OAuth 2.1 authorization server, as an issuer identifier. `fromSupabaseUrl('https://<ref>.supabase.co')` is a small adapter that appends `/auth/v1`.

Both default to values derived from the request as it arrives through the Edge Functions proxy, so **Edge Functions stays zero-config**. Nested and `pipeline()` forms both work, and the config argument remains optional.

```ts
// Edge Functions — unchanged
withOAuthProtectedResource(withSupabase({ auth: 'user' }, handler))

// Any other backend
withOAuthProtectedResource(
  {
    resourceServer: (req) => new URL(req.url).origin + '/api/mcp',
    authorizationServer: fromSupabaseUrl('https://abc123.supabase.co'),
  },
  handler,
)

// Any OAuth server
withOAuthProtectedResource(
  { 
    resourceServer: 'https://api.example.com/mcp', 
    authorizationServer: 'https://example.clerk.accounts.dev' },
  handler,
)
```

### Derivation changes for the Edge Functions default

- Prefers `SUPABASE_PUBLIC_URL` for the origin, then the forwarded headers. `SUPABASE_URL` is never consulted. it's an internal gateway hostname locally (`kong:8000`) and self-hosted (`api-gw:8000`), and on hosted it's pinned to the ref domain even when a custom domain is active. There's a regression test asserting the internal hostname can't leak into the advertised URL.
- Uses `SUPABASE_FUNCTION_SLUG` when the runtime provides it, giving a canonical `/functions/v1/{slug}` with no path parsing. Progressive enhancement — absent, it falls back to reconstructing from the request path, so nothing depends on the rollout.
- The advertised `resource` is now derived so that it always equals the URL the client called, at any path depth.

### Fixes

- Double port suffix when the forwarded host already carries a port.
- Case-sensitive proto comparison (`HTTPS` skipped standard-port handling).
- Comma-joined forwarded values are no longer split-and-guessed. No gateway in front of Edge Functions comma joins these headers, verified across five (hosted relay, self-hosted Envoy, self-hosted Kong, both CLI proxies).

### DIFF
--- a/src/oauth-protected-resource/url.ts
+++ b/src/oauth-protected-resource/url.ts
@@ -1,63 +1,212 @@
+import { getEnv } from '@supabase/middleware'
+
+import {
+  Errors,
+  MissingAuthorizationServerError,
+  MissingResourceServerError,
+} from '../errors.js'
+import { AUTH_PATH_PREFIX, EDGE_FUNCTIONS_PATH_PREFIX } from './paths.js'
+import { isEdgeFunctions } from './runtime.js'
+
 /**
- * Constructs the external-facing base URL from the request,
- * considering the `X-Forwarded-*` headers set by the Supabase Edge Functions proxy.
+ * A configured URL: either a fixed value, or derived per request.
+ *
+ * @category Types
+ */
+export type UrlOption = string | ((req: Request) => string)
+
+/**
+ * Trailing path segment where a resource serves its OAuth Protected Resource
+ * Metadata document (RFC 9728), as a stripping pattern.
  *
  * @internal
  */
-export function getBaseUrl(req: Request): string {
+const METADATA_SUFFIX_PATTERN = /\/oauth-protected-resource$/
+
+/** Strips a trailing slash so URL concatenation doesn't produce `//`. @internal */
+export function trimTrailingSlash(value: string): string {
+  return value.endsWith('/') ? value.slice(0, -1) : value
+}
+
+/**
+ * The externally-visible origin of a Supabase Edge Function.
+ *
+ * `SUPABASE_PUBLIC_URL` wins when set. Otherwise the origin is assembled from
+ * `X-Forwarded-Host`, `-Proto` and `-Port`, each falling back to the request
+ * URL:
+ *
+ * - The port comes from `X-Forwarded-Port` only when the host does not already
+ *   carry one.
+ * - A port that is standard for the scheme (443 on https, 80 on http) is
+ *   omitted.
+ * - The scheme is lowercased.
+ * - Header values are used as-is, never comma-split.
+ *
+ * `SUPABASE_URL` is not consulted — see {@link defaultAuthorizationServer}.
+ *
+ * @internal
+ */
+function edgeOrigin(req: Request): string {
+  const publicUrl = getEnv('SUPABASE_PUBLIC_URL')
+  if (publicUrl) return trimTrailingSlash(publicUrl)
+
   const url = new URL(req.url)
   const host = req.headers.get('X-Forwarded-Host') ?? url.hostname
-  const proto =
+  const proto = (
     req.headers.get('X-Forwarded-Proto') ?? url.protocol.replace(':', '')
-  const port = req.headers.get('X-Forwarded-Port') ?? url.port
+  ).toLowerCase()
+  const port = host.includes(':')
+    ? ''
+    : (req.headers.get('X-Forwarded-Port') ?? url.port)
 
   const isStandardPort =
     (proto === 'https' && port === '443') || (proto === 'http' && port === '80')
 
-  const portSuffix = port && !isStandardPort ? `:${port}` : ''
-
-  return `${proto}://${host}${portSuffix}`
+  return `${proto}://${host}${port && !isStandardPort ? `:${port}` : ''}`
 }
 
 /**
- * Detects the edge function name from the request path.
- * The Supabase proxy strips `/functions/v1` but keeps the function name
- * as the first path segment (e.g. `/my-fn/...` -> function name is `"my-fn"`).
+ * The external path of a Supabase Edge Function.
+ *
+ * With `SUPABASE_FUNCTION_SLUG` set, `/functions/v1/{slug}` — canonical, and
+ * independent of the path the request arrived on. Without it, the received path
+ * minus the metadata suffix, with the `/functions/v1` prefix restored (every
+ * gateway fronting Edge Functions strips that prefix).
+ *
+ * The two forms differ only for a request at a sub-path of the function: the
+ * canonical one reports the function, the reconstructed one the sub-path.
  *
  * @internal
  */
-export function inferFunctionName(req: Request): string | undefined {
-  const url = new URL(req.url)
-  const segments = url.pathname.split('/').filter(Boolean)
-  return segments[0]
+function edgeResourcePath(req: Request): string {
+  const slug = getEnv('SUPABASE_FUNCTION_SLUG')
+  if (slug) return `${EDGE_FUNCTIONS_PATH_PREFIX}/${slug}`
+
+  const received = new URL(req.url).pathname.replace(
+    METADATA_SUFFIX_PATTERN,
+    '',
+  )
+  return `${EDGE_FUNCTIONS_PATH_PREFIX}${received}`
 }
 
 /**
- * Constructs the external-facing URL of the protected resource (the edge function itself).
- * Restores the `/functions/v1` prefix stripped by the Supabase proxy.
+ * The resource identifier to advertise when none was configured: on Edge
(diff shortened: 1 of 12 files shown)


src/oauth-protected-resource/with-oauth-protected-resource.ts
```diff
@@ -11,8 +19,66 @@ export interface OAuthProtectedResourceContribution {
...
+   * proxy). Pass {@link requestUrl} for any backend where the handler receives
+   * the request at its real, externally-visible URL.
+   */
+  urlStrategy?: ResourceUrlStrategy
+  /**
+   * This project's Supabase Auth URL (the bare project URL, e.g.
+   * `https://<ref>.supabase.co` — the same value passed to `createClient()`).
+   * Required off Edge Functions, where it can't be inferred from the request.
+   * Falls back to `SUPABASE_URL` from the environment when omitted. On Edge
+   * Functions this is derived from the request and needs no override.
```
gregnr:
```suggestion
   * The Supabase project base URL (i.e. `https://<ref>.supabase.co`).
   * Required off Edge Functions, where it can't be inferred from the request.
   * Falls back to `SUPABASE_URL` from the environment when omitted. On Edge
   * Functions this is derived from the request and needs no override.
```

src/oauth-protected-resource/url.ts
```diff
@@ -1,24 +1,144 @@
...
+  /**
+   * This project's Supabase Auth base URL, when the strategy can derive it from
+   * the request. Present only for strategies whose origin *is* the Supabase
+   * project's origin (i.e. {@link edgeFunctionProxy}); absent for backends where
+   * the app's own origin is unrelated to the Supabase project, in which case the
+   * caller must supply it explicitly (see `withOAuthProtectedResource`'s
+   * `supabaseUrl`).
+   */
+  defaultAuthorizationServer?(req: Request): string
+}
```
gregnr: This complexity is one of the reasons why I'd push us to introduce an injected `SUPABASE_FUNCTION_SLUG` (or similar) env var within edge runtime instead.`ResourceUrlStrategy` feels like an overly abstract solution that only really exists to deal with edge function proxy quirks. All of the `X-Forwarded-` complexity also goes away with this.

e.g. on edge runtime, logic can reduce to simply:

```ts
const supabaseUrl = getEnvVar('SUPABASE_URL');
const functionSlug = getEnvVar('SUPABASE_FUNCTION_SLUG')
const resourceServerUrl = `${supabaseUrl}${EDGE_FUNCTIONS_PATH_PREFIX}/${functionSlug)
const resourceMetadataUrl = `${functionUrl}/oauth-protected-resource`
```

on other runtimes, we provide a `resourceServer` prop and they fill this out themselves. Something like that. We could even make the prop optionally dynamic if they wanted to build it based on the request url. So generally:

```ts
// no args defaults to the resourceServer generated above on edge runtime, error otherwise
withOAuthProtectedResource()
```
```ts
// set a static url
withOAuthProtectedResource({
  resourceServer: '<some-static-url>'
})
```
```ts
// dynamically generate the url based on the request url
withOAuthProtectedResource({
  resourceServer: (req) => `${req.origin}/...`
})
```

gregnr (commented):
See below inline comment about simplifying this logic with a `SUPABASE_FUNCTION_SLUG` (or similar) env var. I'll also add that accepting a `supabaseUrl` couples this middleware closely to Supabase. Which is fine if that's the goal, but if we want `withOAuthProtectedResource()` to truly be platform agnostic (beyond runtime - auth as well), we could make the SDK more adapter based.

e.g.

```ts
withOAuthProtectedResource({
  authorizationServer: '<any-oauth-server-you-want>'
})
```

`authorizationServer` defaults to `https://<ref>.supabase.co/auth/v1` on edge runtime, errors otherwise. If someone using CF workers wants to use supabase auth, they use an adapter:

```ts
withOAuthProtectedResource({
  authorizationServer: fromSupabaseUrl(`https://<ref>.supabase.co`)
})
```

and under the hood, we determine the authorization server metadata url from the corresponding well-known endpoint, e.g. 
`https://<ref>.supabase.co/auth/v1` -> `https://<ref>.supabase.co/.well-known/oauth-authorization-server/auth/v1`

But this gives them the freedom to use any authorization server, e.g. clerk.

---

So to summarize I'm suggesting we reconsider the config options to accept `resourceServer` and `authorizationServer` (final names TBD) with defaults and adapters instead of `urlStategy` and `supabaseUrl`.

mandarini:
Nice! :D I like the precedence tests (especially the `kong:8000` internal-hostname regression)!

One "should-fix": can we use the engine's `getEnv` instead of our internal `getEnvVar` for `SUPABASE_PUBLIC_URL`/`SUPABASE_URL`. That's Rule 2 of the [authoring guide](https://github.com/supabase/middleware/blob/main/docs/authoring-guide.md) (we should never touch `Deno.env`/`process.env` directly). The reason is `getEnv` reads Workers bindings that #116 just made flow correctly, so the env fallbacks would work off Deno/Node. I think it's just a small change in `url.ts` + `resolveAuthServer`.

One small question: the metadata route went from exact `/{fn}/oauth-protected-resource` to any path ending in `/oauth-protected-resource`. I think suffix matching is defensible, but it shadows app routes with that suffix at any depth, and on the edge strategy a nested metadata GET reports the top-level function as the resource. Can you confirm it's intended and maybe we could add a doc line saying so?

raulb:
@gregnr I think you're right. The more I check the different permutations (different environments), the more I think we should lean more towards being deterministic and stop doing so many forks by checking headers and stripping paths. This is looking way too hacky considering we do own the lifecycle of these environment variables being set. I'll bounce this idea off the functions team. Here's [the commit](https://github.com/supabase/server/pull/117/commits/3f152214237728852e9dfa90be97a045c7af4cfe) with the refactor. 

@mandarini I think I also addressed all your feedback. Please take another look 🙏

mandarini (changes requested):
Soo... requesting changes, but it's all small 😅😅

1. The Clerk example adds third-party issuer support. AI-1047 says the auth server stays Supabase Auth only, per the MCP Endpoints PRFAQ. I think the direction makes sense, but can we confirm with product before we ship and document it? If it's fine, a note on the ticket is enough.

2. Off Edge Functions with no `authorizationServer`, the default derives `{origin}/auth/v1` from the request. That points at the app's own origin, so it's wrong on Vercel, Workers, and everywhere else off platform. Can we either throw there or fall back to `SUPABASE_URL`, which was the ticket's original plan? A silently wrong default is the one option we shouldn't keep, imo.

3. Can we please add a test for a 401 from the inner handler with `resourceServer` configured? That's the main off-platform flow (401, then discovery via `WWW-Authenticate`), and right now only the metadata GET route is tested with config.

4. Docs: the shadowing note from my earlier comment (any GET path ending in `/oauth-protected-resource` never reaches the inner handler), `fromSupabaseUrl` in the README export table, and `SUPABASE_PUBLIC_URL` / `SUPABASE_FUNCTION_SLUG` in `docs/environment-variables.md`.

Everything else looks good!! Thanks @raulb !!

raulb:
@mandarini 

> 1. The Clerk example adds third-party issuer support. AI-1047 says the auth server stays Supabase Auth only, per the MCP Endpoints PRFAQ. I think the direction makes sense, but can we confirm with product before we ship and document it? If it's fine, a note on the ticket is enough.

Yeah, PRFAQ didn't scope the changes on different authentication servers, but both @gregnr and I considered good to at least design this middleware keeping that in mind. I'll update the issue.

Good callout on 2-4! Fixed the default value, and I added more documentation and validation to throw descriptive errors when those values are not in place. Let me know. I appreciate you looking at it 🙏

mandarini (approved):
perfect, thanks!

raulb:
Re-verified once more. Everything works. Shipping.

PR #130, opened 2026-08-27
feat(errors): add self-identifying errors with hints and diagnostics

johnstonmatt:
Nearly every failure used to return the same generic `{ message: "Invalid credentials", code: "INVALID_CREDENTIALS" }`, naming neither the cause nor the source library. All errors now identify themselves and explain what to do next.

- Introduce `SupabaseServerError`, a shared base carrying `source`, a `[@supabase/server]` message prefix, a `docs` link, an optional `hint`, and non-sensitive `details`; add a single `errorResponse()` helper that renders it everywhere (with an `x-supabase-server-error` header exposed via CORS), while keeping top-level `message`/`code` unchanged for existing consumers
- Make `verifyUserJwt` and the auth mode chain report the specific reason a request failed instead of a generic invalid-credentials fallback, adding `MISSING_CREDENTIALS`, `INVALID_API_KEY`, `INVALID_JWT`, `JWKS_NOT_CONFIGURED`, `JWKS_FETCH_FAILED`, and `NO_KEYS_CONFIGURED`, and surfacing server misconfiguration as 500 instead of 401
- Give `withClaims`, `withRequiredClaims`, `withPostgresClient`, and `withPostgresAdminClient` the same specific, hinted errors instead of hand-rolled bodies
- Add `errors: { detailed: false }` to `withSupabase` to trim response bodies to just `code` and `message` while leaving status, the error header, and the in-process error object untouched
- Split `UNUSABLE_CREDENTIAL` out from `MISSING_CREDENTIALS` so a credential that arrived but couldn't be used (wrong kind or unreadable `Authorization` header) is distinguishable from one that never arrived, via a shared `diagnoseAuthorizationHeader` classifier used by both `verifyAuth` and the `withRequiredClaims` gate
- Ensure `details` never carries secret material — API keys are reported by prefix format and named keys by name only
- Rewrite `docs/error-handling.md` and update `docs/api-reference.md` / `docs/postgres.md` to document the new error classes, codes, and response-trimming behavior


docs/api-reference.md
```diff
@@ -550,40 +567,105 @@ import type {
...
+  hint?: string
+  details?: Record<string, unknown>
+  docs?: string // overrides the generated URL
+  cause?: unknown
 }
 ```
 
 ---
 
 ## Error Code Constants
```
tomaspozo, line 644: Maybe we could add a "Possible Causes" or similar column to this table? So is easier for agents/users to think of a root case. Does not need to be very detailed, but for example, something like: "apikey or Authorization headers are missing"

johnstonmatt: we can do that! do you think it is needed in addition to the other places this PR adds info, like [docs/error-handling.md](https://github.com/supabase/server/blob/FUNC/improve-withSupabase-errors/docs/error-handling.md) ?

src/middleware/required-claims/index.ts
```diff
@@ -92,29 +100,42 @@ export const withRequiredClaims: Middleware<
     // header — they are API keys, not user JWTs, so they cannot pass a gate
     // that requires verified user claims.
     if (!token || token.startsWith('sb_')) {
-      return Response.json(
-        { message: 'Invalid credentials', code: InvalidCredentialsError },
-        { status: 401 },
+      const { apikey } = extractCredentials(req)
+      return errorResponse(
+        Errors[MissingCredentialsError]({
```
mandarini: question: on this branch (and the matching one in `verifyCredentials`), the top-level code is `MISSING_CREDENTIALS` even though a credential did arrive, just the wrong kind. The `received.authorization: 'api-key'` details and the "API keys belong in the apikey header" hint cover the diagnosis. But with `errors: { detailed: false }` both get stripped, and the caller sees a bare `MISSING_CREDENTIALS` while they are definitely sending a key. Is one fallback code the deliberate trade-off to keep the catalogue small, or is "credential of the wrong kind" worth its own code? Not blocking either way. The hint is what matters for the Studio case!

johnstonmatt: it was an oversight on my part, we should have states that are uniquely identifiable by code, good catch! Fixed in 21dc29c.

What do you think about `{ detailed: false }` more generally? I was worried some users might consider the verbose errors over-sharing, maybe even to a degree where it is perceived insecure. I did bundle it in a separate commit though, in case that fear is imagined and we don't want to maintain that config/code-path

mandarini: I say we keep it. Default true and the "verbosity control, not a security boundary" framing are both right, I think!

mandarini:
@johnstonmatt One follow-up, since `detailed: false` makes the code the only signal: `supabase-js` sends the publishable key in both apikey and `Authorization: Bearer`. On an `auth: 'user'` endpoint that request hits the `apikey !== 'absent'` branch in `explainFallthrough` first and returns `INVALID_API_KEY` ("check you're pointing at the right project"), never `UNUSABLE_CREDENTIAL`. The new tests only set the Authorization header. Could the key-in-Authorization check win whenever the mode list accepts no API keys?

src/errors.ts
```diff
@@ -203,25 +557,459 @@ export const CreateSupabaseClientError = 'CREATE_SUPABASE_CLIENT_ERROR'
...
+        details: context
+          ? {
+              acceptedAuthModes: context.authModes,
+              received: context.received,
+            }
+          : undefined,
+      },
+    ),
+
+  [CreateSupabaseClientError]: (options?: { cause?: unknown }): AuthError =>
```
mandarini, line 927: Small observation: The two middleware call sites that raise this (`src/middleware/client/index.ts:52`, `src/middleware/admin-client/index.ts:48`) still call `Errors[CreateSupabaseClientError]()` with no arguments, so `error.cause` is undefined there while the hint tells the reader to log it. `create-supabase-context.ts:92` passes `{ cause: e }`. I think these two should match.

docs/error-handling.md
```diff
@@ -1,64 +1,280 @@
...
+| [`MISSING_SUPABASE_URL`](#missing_supabase_url)                       | `SUPABASE_URL` is not set                                          |
+| [`MISSING_PUBLISHABLE_KEY`](#missing_publishable_key)                 | Named publishable key not found in `SUPABASE_PUBLISHABLE_KEYS`     |
+| [`MISSING_DEFAULT_PUBLISHABLE_KEY`](#missing_default_publishable_key) | No default publishable key found                                   |
+| [`MISSING_SECRET_KEY`](#missing_secret_key)                           | Named secret key not found in `SUPABASE_SECRET_KEYS`               |
+| [`MISSING_DEFAULT_SECRET_KEY`](#missing_default_secret_key)           | No default secret key found                                        |
+| [`MISSING_RESOURCE_SERVER`](#missing_resource_server)                 | `withOAuthProtectedResource` cannot derive a `resourceServer`      |
+| [`MISSING_AUTHORIZATION_SERVER`](#missing_authorization_server)       | `withOAuthProtectedResource` cannot derive an authorization server |
+| [`ENV_ERROR`](#env_error)                                             | Generic environment error                                          |
```
mandarini, line 222: `MISSING_CONNECTION_STRING` is missing from this table and has no `###` section below, so the `error.docs` URL that `docsFor()` generates for `withPostgresClient` / `withPostgresAdminClient` points at an anchor that doesn't exist. It's also absent from the api-reference constants table.

docs/error-handling.md
```diff
@@ -1,64 +1,280 @@
...
+
+`details.jwt` carries the token's `alg` and `kid` — both client-supplied and public — which is what you need to debug a JWKS mismatch. Claim values are never included.
+
+A present-but-invalid JWT rejects immediately rather than falling through to the next auth mode, so this code always wins over a later mode's failure.
+
+### `INVALID_CREDENTIALS`
+
+Fallback code, returned when a credential was present but no more specific code applies.
+
+> **Changed in v1.6.** This used to be the only code returned for a failed request. The specific codes above now cover essentially every real failure, so match on those instead. `INVALID_CREDENTIALS` and `Errors[InvalidCredentialsError]()` remain exported and working.
```
mandarini: I think this should be version `1.5`

mandarini (approved):

PR #133, opened 2026-08-31
chore: force release 1.5.1

mandarini:
Release-As: 1.5.1

PR #135, opened 2026-08-31
chore: rewind phantom 1.5.0 version files

mandarini:
Release-As: 1.5.1


mandarini:
Context for reviewers, since the release keeps failing in non-obvious ways:

- The Aug 25 release run created the `server-v1.5.0` tag and GitHub release before the JSR publish failed. #120 deleted both.
- GitHub immutable releases reserve a published tag name forever. Deleting the tag and release does not free the name. `server-v1.5.0` can never exist again.
- That is why the release step failed again after #121 merged: release-please tried to recreate the same tag and got "tag_name was used by an immutable release".
- So we skip 1.5.0 and ship 1.5.1 instead. #133 added the `Release-As: 1.5.1` override.
- Main still says 1.5.0 was released (package.json, jsr.json, manifest, CHANGELOG) while no `server-v1.5.0` tag exists. release-please cannot resolve that version to a tag, loses its diff anchor, and rebuilt #134 with the entire repo history in the changelog.
- This PR rewinds the version files to 1.4.1, the last version with a real tag. Same move as #120. release-please then regenerates #134 anchored at `server-v1.4.1`, targeting 1.5.1.
- 1.5.0 never reached npm or JSR, so skipping it affects no one.

spydon (approved):

PR #154, opened 2026-09-10
docs: add MCP server guide and surface the OAuth entry point in the skill

raulb:
Part of AI-1009

Are documentations specifically about `withOAuthProtectedResource()` in the context of using it for MCP servers, clarifying its usage with Supabase Edge Functions or not.

**Note:** I also updated the library components usage here https://github.com/supabase/supabase/pull/50215 which will ship as part of https://github.com/supabase/supabase/pull/49579.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -58,13 +58,13 @@ Order matters.
 
 `withOAuthProtectedResource()` runs **before** the auth gate, so it sees unauthenticated requests. It does two things:
 
-| Request                                       | Response                                                                                        |
-| --------------------------------------------- | ----------------------------------------------------------------------------------------------- |
-| `GET {resource}/oauth-protected-resource`     | RFC 9728 Protected Resource Metadata (`resource`, `authorization_servers`)                      |
-| `OPTIONS {resource}/oauth-protected-resource` | `204` with permissive CORS headers, so browser-based clients can read the document cross-origin |
-| Any response from below with status `401`     | Adds `WWW-Authenticate: Bearer resource_metadata="…"` unless the handler already set one        |
+| Request                                       | Response                                                                                               |
+| --------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
+| `GET {resource}/oauth-protected-resource`     | RFC 9728 Protected Resource Metadata (`resource`, `authorization_servers`, `bearer_methods_supported`) |
+| `OPTIONS {resource}/oauth-protected-resource` | `204` with permissive CORS headers, so browser-based clients can read the document cross-origin        |
+| Any response from below with status `401`     | Adds `WWW-Authenticate: Bearer resource_metadata="…"` unless the handler already set one               |
 
-That header is how a client that hit a `401` finds the metadata, and through it the authorization server, without guessing URLs. It is generic OAuth middleware; nothing in it is MCP-specific.
+That header is how a client that hit a `401` finds the metadata, and through it the authorization server, without guessing URLs. It is generic OAuth middleware; the only MCP-specific detail is that the preflight allows the `mcp-protocol-version` request header.
 
 `withSupabase({ auth: 'user' })` is the gate. Requests without a valid user JWT get a `401` (which the entry above enriches); requests with one reach the handler with `ctx.supabase` scoped to that user, so RLS applies to everything the tools do.
 
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -1,0 +1,116 @@
+# MCP servers
+
+> **Alpha.** `withOAuthProtectedResource` and composing `withSupabase` as a
+> `pipeline` entry track `@supabase/middleware` 0.x. The config shape, the
+> contributed context key, and the metadata route may change in a minor
+> release. The nested `withOAuthProtectedResource(withSupabase(config, handler))`
+> form is stable.
+
+An MCP server your app exposes to its users is an HTTP endpoint that MCP clients (Claude, ChatGPT, Cursor, VS Code, Claude Code) call after an OAuth 2.1 flow. `@supabase/server` covers the two Supabase-specific parts: OAuth discovery for the client, and turning the user's token into an RLS-scoped Supabase client for your tools. The MCP transport and the tools come from any MCP library.
+
+```ts
+import { pipeline } from '@supabase/middleware'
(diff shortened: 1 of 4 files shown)


mandarini:
A few additions from running this on a real hosted Edge Function (diabuddy, `@supabase/server` 1.6.0, Deno 2.9). Each would fit in one sentence:

- **Typed client for writes.** `withSupabase({ auth: 'user' })` without a generic gives `SupabaseClient<unknown>`. The sample only selects, but a tool that writes fails typecheck: `insert({ title })` resolves the argument to `never[]`. Showing `withSupabase<Database>` (or `<any>`) in the sample avoids the first support question.
- **Factory errors are invisible.** An exception thrown inside the `createMcpHandler` factory (a duplicate tool name, a failed schema fetch) comes back as a 500 with nothing in the function logs. A `try/catch` with `console.error` in the factory is worth showing.
- **OPTIONS on the metadata route.** The entry also answers the preflight for `{resource}/oauth-protected-resource` with CORS headers. A third row in the table would document it for browser-based clients.
- **The limits are the handler's, not the library's.** The one-request, one-response shape comes from building a `McpServer` per request with `createMcpHandler`; `@supabase/server` does not constrain the transport. Saying so avoids the impression that the entries rule out a stateful transport.
- **Root path without a slug.** On a CLI older than 2.117 serving the function at `/`, derivation throws `MISSING_RESOURCE_SERVER`. One line here saves a support question.

Verified against 1.6.0: the inline-handler TS2345 reproduces exactly as described, and the `(req, ctx) => handleMcp(req, ctx)` form compiles.

docs/mcp.md
```diff
@@ -0,0 +1,96 @@
...
+
+That header is how a client that hit a `401` finds the metadata, and through it the authorization server, without guessing URLs. It is generic OAuth middleware; nothing in it is MCP-specific.
+
+`withSupabase({ auth: 'user' })` is the gate. Requests without a valid user JWT get a `401` (which the entry above enriches); requests with one reach the handler with `ctx.supabase` scoped to that user, so RLS applies to everything the tools do.
+
+The handler is passed inline. Passing a separately declared function typed `(req: Request, ctx: SupabaseContext) => Promise<Response>` makes TypeScript collapse the inferred context to `object` (TS2345); wrap it as `(req, ctx) => handleMcp(req, ctx)` instead.
+
+## URLs
+
+On Supabase Edge Functions the metadata is derived with no configuration: the public origin from the gateway's `X-Forwarded-*` headers (or `SUPABASE_PUBLIC_URL`), the path from `SUPABASE_FUNCTION_SLUG` with the `/functions/v1` prefix restored, and the issuer as `{origin}/auth/v1`. This works locally and hosted; Supabase CLI 2.117.0 or later injects the slug and public URL locally.
```
mandarini: supabase/cli#6345 (in 2.117.0) injects `SUPABASE_FUNCTION_SLUG` only. `SUPABASE_PUBLIC_URL` is honored when set (it wins in `edgeOrigin`), but the CLI does not set it. Suggest "Supabase CLI 2.117.0 or later injects the slug locally."

docs/mcp.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+- `resourceServer`: the public URL of this endpoint. Required off Edge Functions; `MissingResourceServerError` (`MISSING_RESOURCE_SERVER`) otherwise.
+- `authorizationServer`: the OAuth issuer. Falls back to `SUPABASE_PUBLIC_URL`, then `SUPABASE_URL`, each with `/auth/v1`; `MissingAuthorizationServerError` (`MISSING_AUTHORIZATION_SERVER`) if neither is set. `fromSupabaseUrl(projectUrl)` builds it from a project URL. A non-Supabase OAuth 2.1 server (Clerk, WorkOS, Auth0) works too.
+
+Both accept a string or `(req: Request) => string`.
+
+## Supabase Auth prerequisites
+
+The project must have the OAuth 2.1 server enabled, dynamic client registration on (MCP clients register themselves), an asymmetric signing key (ES256 or RS256; OAuth 2.1 tokens cannot be HS256), and a consent screen hosted in the app's frontend. On Edge Functions set `verify_jwt = false` for the function in `config.toml`; the gate does the verification, and the gateway must let the unauthenticated discovery request through.
```
mandarini: "OAuth 2.1 tokens cannot be HS256" is not quite the reason. Supabase Auth still issues HS256 access tokens on projects using the legacy secret; only OIDC ID tokens require an asymmetric key. The constraint is this library: `withSupabase({ auth: 'user' })` verifies user JWTs against the project JWKS and rejects HS256 (the "Legacy keys and JWTs are not supported" note in `docs/auth-modes.md`). Suggest: "an asymmetric signing key (ES256 or RS256), because `withSupabase` verifies user JWTs against the project JWKS and does not accept HS256".

docs/mcp.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Limits
+
+Streamable HTTP in its simplest form: one request in, one response out. No server-initiated stream, so no MCP sampling. Elicitations that need an open channel are not available either.
+
+## See also
+
+- [Give your app an MCP server](https://supabase.com/docs/guides/ai-tools/byo-mcp) on supabase.com, the end-to-end guide
+- [MCP Server block](https://supabase.com/library/docs/headless/mcp-server), an installable Edge Function built on this
+- [`docs/api-reference.md`](api-reference.md) for the full `OAuthProtectedResourceConfig`
```
mandarini: `docs/api-reference.md` on main only carries the two `MISSING_*` error rows; `OAuthProtectedResourceConfig` (`resourceServer`, `authorizationServer`, `UrlOption`) is not documented there yet. Either add the config table in this PR or point at the type's JSDoc.

mandarini (commented):
Checked the page against `origin/main` (1.6.0), the URL derivation in `src/oauth-protected-resource/url.ts`, and a hosted Edge Function running this stack. Three factual fixes inline. The additions worth considering are in a separate comment.

raulb:
@mandarini thank you for the feedback! You were right on all three. I'm currently running tests following across the library blocks https://github.com/supabase/supabase/pull/50215 and future docs https://github.com/supabase/supabase/pull/50218, so I'm hoping to leave this PR ready for review today.

mandarini (approved):

PR #160, opened 2026-09-12
feat: validate jwt audience and issuer

mandar1045:
## TL;DR

Validate the JWT `aud` and `iss` claims during user token verification through config overrides.

## Problem

Token verification currently doesn't validate the audience or issuer claims. This means a valid token issued for one service could potentially be accepted by another service.

## Solution

Adds optional `audience` and `issuer` config overrides to `verifyCredentials`, `withClaims`, and `withRequiredClaims`.

This works with both symmetric and asymmetric keys, and doesn't read from any reserved environment variables.

## Ref

* Supersedes #41


src/core/resolve-env.ts
```diff
@@ -181,6 +181,8 @@ export function resolveEnv(
       overrides?.secretKeys ??
       resolveKeys('SUPABASE_SECRET_KEY', 'SUPABASE_SECRET_KEYS'),
     jwks: overrides?.jwks ?? resolveJwks(),
+    audience: overrides?.audience ?? null,
+    issuer: overrides?.issuer ?? null,
```
kallebysantos: As discussed in the related PR https://github.com/supabase/server/pull/41, we shouldn't reference these options as env / `resolveEnv()`

kallebysantos (changes requested):

src/core/resolve-env.ts
mandar1045: Yep, @kallebysantos! Since these aren’t env vars, I don’t think they belong in SupabaseEnv or resolveEnv().

I’ll take them out of resolveEnv() and add audience and issuer directly to WithSupabaseConfig and VerifyCredentialsOptions, similar to withClaims.

Would you rather keep them as top-level options (withSupabase({ auth: 'user', audience, issuer })) or put them under a jwt config ?

mandar1045: Updated! Removed `audience` and `issuer` from `SupabaseEnv` and `resolveEnv()`, and added them directly to `WithSupabaseConfig` and `VerifyCredentialsOptions`, similar to `withClaims`.

mandarini (changes requested):
Thanks for picking this up and for moving the options off `resolveEnv`. The logic is right and every call site gets the options. Three small changes before merge.

- Drop `| null` from the new field types and pass the options straight to `jwtVerify` (or guard with `!== undefined`). With the current truthiness check, an empty string turns validation off with no warning and an empty array rejects every token.
- Fix the example in `docs/security.md`. Supabase Auth sets `aud` to `authenticated` and `iss` to `https://<ref>.supabase.co/auth/v1`, so `audience: 'my-service'` would reject every Supabase token. Use `issuer: fromSupabaseUrl(SUPABASE_URL)` as the example and point to `withRequiredClaims` rather than `withClaims`, since `withClaims` lets requests with no token through.
- Add a one-line JSDoc to each new field, matching the fields around them. Mention that they apply to `user` mode only.

On your question about grouping: top level is fine. It matches where `jwks` sits on `withClaims`.

src/core/verify-user-jwt.ts
```diff
@@ -248,6 +256,11 @@ export async function verifyUserJwt(
     const jwkResolver = getJwksResolver(jwks)
     let payload: JWTPayload | null = null
 
+    const verifyOptions: JWTVerifyOptions = {
```
mandarini, line 290: `audience: ''` / `issuer: ''` silently disables the check. `jose` `6.2.0` matches these with a truthiness test (`if (audience && ...)`), not `!== undefined`, so an empty string (e.g. from an unset env var) is accepted as "configured" but validates nothing. Guard against empty string the same way you guarded `| null`.

mandar1045: Thanks for catching that. Added guards for empty strings and updated the tests.

src/core/verify-user-jwt.ts
```diff
@@ -244,10 +252,30 @@ export async function verifyUserJwt(
...
+  ) {
+    throw new Error('JWT audience option cannot be empty')
+  }
+  if (
+    options?.issuer === '' ||
+    (options?.issuer as unknown) === null ||
+    (Array.isArray(options?.issuer) && options.issuer.some((i) => !i))
+  ) {
+    throw new Error('JWT issuer option cannot be empty')
+  }
```
mandarini, line 284: Thanks for closing the empty-string hole, but this throws, and nothing else in this function does. `verifyUserJwt` always returns `{ ok: false, failure }` on failure; every caller (`withClaims`, `withRequiredClaims`, `gate.ts`, `createSupabaseContext`) turns that into a `Response`. None of them catch a thrown `Error`, so an empty `audience`/`issuer` now crashes the request instead
of returning a 4xx.

Return the failure instead of throwing, same shape as the checks above it:

```
    return {
      ok: false,
      failure: {
        kind: 'token',
        reason: 'the configured "audience" option is empty',
        hint: 'Pass a non-empty string or array, or omit the option to skip audience validation.',
        jwt,
      },
    }
```

(mirror for `issuer`). No other files need to change, this flows through the existing `errorResponse` mapping in each caller already.

mandar1045: Updated! Switched to returning the failure result instead of throwing, and updated the tests to assert the 401 response and reason.

mandarini (approved):

kallebysantos (approved):

PR #166, opened 2026-09-15
docs: add Deno PR-build and pg Date notes for consumers

mandarini:
Three consumer-facing notes for Deno and Edge Functions users, each placed where its audience already looks. Docs only, no runtime change.

- `CONTRIBUTING.md`: a "Trying a PR build" subsection. pkg.pr.new for Node, and the esm.sh route for Deno, since `npm:` rejects tarball URLs and an import map onto an extracted `dist/index.mjs` passes `deno check` without checking anything. Pin the commit sha and pass `deps` for `@supabase/supabase-js`. Ends with the `--minimum-dependency-age 0` flag for same-day npm releases.
- `docs/postgres.md`: a "Dates come back as `Date`" subsection. `pg` returns `date`, `timestamp`, and `timestamptz` as `Date` objects. Declare them as `Date` in the row type, or cast `::text` in SQL when the value feeds a PostgREST filter or a JSON body.
- `src/core/postgres-pool.ts`: the same `Date` point added to the `query` tsdoc, so TypeDoc carries it into the generated reference.

The Deno optional-peer note for `pg` itself lives on the docs site (supabase/supabase#50412), next to the install commands.


docs/postgres.md
```diff
@@ -69,6 +69,17 @@ const rows: NoteRow[] = await ctx.postgres.query`select id, body from notes`
 
 Passing a plain string to `query` throws. The two calls differ only in their brackets, so refusing is safer than reinterpreting one as the other.
 
+### Dates come back as `Date`
+
+`pg` returns `date`, `timestamp`, and `timestamptz` columns as JavaScript `Date` objects, not strings. Declare those fields as `Date` in your row type. When a value feeds a PostgREST filter through `ctx.supabase`, or goes straight into a JSON body, cast it in SQL instead:
```
spydon, line 74: The other markdown seems to have a max line length, but not this one. Maybe something that could be aligned in the future.

spydon (approved):

PR #167, opened 2026-09-15
feat: derive the `JWKS URL` from `SUPABASE_URL`

raulb:

auth: 'user' now works with only `SUPABASE_URL` and `SUPABASE_PUBLISHABLE_KEY` set, and this matches the zero-config experience on Edge Functions.

`withSupabase({ auth: 'user' })` needs a JWKS to verify user tokens, and the library is currently reading it from `SUPABASE_JWKS` or `SUPABASE_JWKS_URL` (for context Edge Functions inject `SUPABASE_JWKS`)

Other runtimes do not like Next.js, TanStack Start or a plain Node server, a request with a valid token was failing with a 500 (code `JWKS_NOT_CONFIGURED`). Without this change, the user needs to manually set `SUPABASE_JWKS_URL` even though every Supabase project publishes its keys at the same path.

## What is the new behavior?

When `SUPABASE_JWKS` and `SUPABASE_JWKS_URL` are both unset, the library derives the JWKS URL from `SUPABASE_URL`.

## Additional context

Found while validating the middleware outside Edge Functions for Linear AI-1048.


mandarini (changes requested):
Thanks Raul!! Some changes that are needed (claude helped me track these down):

**1. The derived URL skips `overrides.url`**

- `resolveEnv` resolves `url` from `overrides?.url ?? getEnvVar('SUPABASE_URL')` ([L196](https://github.com/supabase/server/blob/b1a5e0b7291f5b60297cd4c7aaec7e816d29dcd1/src/core/resolve-env.ts#L196)), but `resolveJwks()` reads `getEnvVar('SUPABASE_URL')` directly ([L140](https://github.com/supabase/server/blob/b1a5e0b7291f5b60297cd4c7aaec7e816d29dcd1/src/core/resolve-env.ts#L140)). The override never reaches the derivation.
- `resolveEnv({ url: 'http://localhost:54321' })` with no `SUPABASE_URL` in the env returns `jwks: null`. That is the exact example in the docstring above it ([L187-L188](https://github.com/supabase/server/blob/b1a5e0b7291f5b60297cd4c7aaec7e816d29dcd1/src/core/resolve-env.ts#L187-L188)).
- Suggestion: pass the resolved `url` into `resolveJwks(url)` and derive from that. Add a test with `overrides.url` set and `SUPABASE_URL` unset.

**2. Self-hosted: `SUPABASE_URL` is the internal gateway**

- The docstring ([L151](https://github.com/supabase/server/blob/b1a5e0b7291f5b60297cd4c7aaec7e816d29dcd1/src/core/resolve-env.ts#L151)) and the error hint ([errors.ts L851-L860](https://github.com/supabase/server/blob/b1a5e0b7291f5b60297cd4c7aaec7e816d29dcd1/src/errors.ts#L851-L860)) name `http://kong:8000` as the case that yields `null`. That is the default self-hosted setup, not an edge case:
  - The compose file sets `SUPABASE_URL: http://api-gw:8000` next to `SUPABASE_PUBLIC_URL` for the functions container ([docker-compose.yml L453-L457](https://github.com/supabase/supabase/blob/232ce7e68cf50969da75254b220c1f439420cea6/docker/docker-compose.yml#L453-L457)) and for studio ([L51-L52](https://github.com/supabase/supabase/blob/232ce7e68cf50969da75254b220c1f439420cea6/docker/docker-compose.yml#L51-L52)). Operators put their real domain in `SUPABASE_PUBLIC_URL` ([.env.example L97](https://github.com/supabase/supabase/blob/232ce7e68cf50969da75254b220c1f439420cea6/docker/.env.example#L97)).
  - The CLI follows the same split: `SUPABASE_URL=http://kong:8000` for local Edge Functions ([serve.ts L1820-L1827](https://github.com/supabase/cli/blob/5a524e9bf0358bf7e857b47c02a44a8d551671c5/apps/cli/src/shared/functions/serve.ts#L1820-L1827), which also injects `SUPABASE_JWKS`, so local Edge is fine) and the `SUPABASE_URL` / `SUPABASE_PUBLIC_URL` pair for studio ([studio.service.ts L84-L85](https://github.com/supabase/cli/blob/5a524e9bf0358bf7e857b47c02a44a8d551671c5/apps/cli/src/commands/start/services/studio.service.ts#L84-L85)).
  - This repo already handles it for the auth issuer: `defaultAuthorizationServer` checks `SUPABASE_PUBLIC_URL` before `SUPABASE_URL` ([url.ts L153-L163](https://github.com/supabase/server/blob/d298878e8df89bc231df3154b9aafac03ffc0730/src/oauth-protected-resource/url.ts#L153-L163)), and the env docs list `SUPABASE_PUBLIC_URL` as the external URL for self-hosted ([environment-variables.md L13](https://github.com/supabase/server/blob/d298878e8df89bc231df3154b9aafac03ffc0730/docs/environment-variables.md#L13)).
- So any stack built from the compose file still gets `jwks: null` after this PR unless it sets `SUPABASE_JWKS_URL` by hand. `SUPABASE_JWKS` is commented out in that compose file ([L453-L454](https://github.com/supabase/supabase/blob/232ce7e68cf50969da75254b220c1f439420cea6/docker/docker-compose.yml#L453-L454)), so self-hosted Edge Functions don't get the inline JWKS the description assumes either.
- Suggestion: derive from `SUPABASE_PUBLIC_URL` first, then `SUPABASE_URL`, same order as `defaultAuthorizationServer`. Hosted projects never set `SUPABASE_PUBLIC_URL`, so cloud is unaffected. Update the resolution order in the docs and the hint, and add a test.
- One caveat for the docs: `SUPABASE_PUBLIC_URL` has to be reachable from where the server runs. The default `http://localhost:8000` passes the loopback rule but points at the container itself.

Happy to pair on either if useful.

mandarini (approved):

PR #168, opened 2026-09-17
docs: add typechecked framework bridges for middleware entries

mandarini:
Adds `examples/frameworks/` with a copyable, typechecked bridge and a minimal app for Hono, H3, Elysia, and NestJS. Each bridge runs a `@supabase/middleware` entry array inside the framework's own middleware slot, folds the pipeline once so entries keep their state across requests, and fails to compile at the call site on duplicate keys or missing prerequisites. A new `typecheck:examples` script runs in CI and lint covers the folder; nothing ships to npm and no dependency is added. Verified with typecheck, lint, the unit suite, negative compile cases, and a runtime script that drives the Hono, H3, and Elysia bridges with real requests. The guide that explains the bridges and the adapter migration is supabase/supabase#50514; tracked in [SDK-1599](https://linear.app/supabase/issue/SDK-1599).

Blocks: https://github.com/supabase/supabase/pull/50514


examples/frameworks/elysia/supabase-middleware.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Elysia plugin that exposes the entries' contributions on the route context.
+ *
+ * `Entries` is type-only. Pass the type of the same tuple `wrapElysia`
+ * receives, so route handlers see the keys that tuple contributes.
+ */
+export function supabaseCtx<const Entries extends readonly AnyEntry[]>() {
+  return new Elysia()
+    .resolve((c) => (HANDOFF.get(c.request) ?? {}) as Contributions<Entries>)
```
tomaspozo, line 33: `supabaseCtx()` and `wrapElysia()` are independent, but only the pair is safe — and `?? {}` makes the unsafe half look like it worked.

`supabaseCtx()` can be mounted on an app that is never passed through `wrapElysia`. When that happens the entries never run, `HANDOFF.get(c.request)` misses, and `?? {}` turns "the pipeline did not run" into "the pipeline ran and contributed nothing" — cast to `Contributions<Entries>`, so the types still claim `jwtClaims` is there.

The two ways this shows up, both verified on this branch (bf89cda, real `withRequiredClaims()`, request with no `Authorization` header):

```
                                       /ping   /me (reads c.jwtClaims.sub)
export default { fetch: wrapElysia() }  401     401     <- app.ts as written
app.listen(3000)                        200     500
export default app  (app.fetch)         200     500
```

So a route that dereferences a contribution fails closed, but with an unhandled `TypeError: undefined is not an object (evaluating 'c.jwtClaims.sub')` rather than anything pointing at the real cause — and **a route that touches no contribution serves unauthenticated**. `/ping` above is gated in the documented setup and wide open in the other two.

Both mistakes are ordinary Elysia. `app.listen()` and `export default app` are how Elysia apps are normally served; `app.ts` uses `export default { fetch: ... }` instead, and nothing in the file, the bridge docstrings, or `examples/frameworks/README.md` says the two must be paired. The README says "Copy it into your project as is", which is exactly when the pairing gets dropped. The other three bridges don't have this shape — `toHono`, `toH3` and `toNestGuard` each expose one function, so there is no half to forget.

Worth noting it is only *fully* unwrapped apps: a sub-app that does `.use(supabaseCtx())` and is mounted into a wrapped parent is fine (verified, 401), since the same `Request` reaches the resolve hook. The hole is when nothing in the chain is wrapped.

`tsc --noEmit -p examples` is clean on an app that mounts `supabaseCtx()` and calls `app.listen()` with no `wrapElysia` in the file, so CI won't catch it either.

Suggestion — make the missing handoff throw instead of defaulting:

```ts
.resolve((c) => {
  const contributions = HANDOFF.get(c.request)
  if (!contributions) {
    throw new Error(
      'supabaseCtx() ran without wrapElysia(): the entries never ran, so ' +
        'this request was not gated. Serve the app through ' +
        '`wrapElysia(entries, (req) => app.handle(req))`.',
    )
  }
  return contributions as Contributions<Entries>
})
```

Verified against the same matrix: the documented setup is unchanged (401 on both routes), and both unwrapped forms now fail closed on every route with that message instead of serving `/ping`. A line in `app.ts` and the bridge docstring saying the two halves are one unit would help too.

via Claude

mandarini: Applied in 2e3c05a552f0eb4fb1267e34a95ed2b5b4a29c50.

Elysia: `supabaseCtx` now throws when the handoff is missing, with your message. The bridge docstrings, `app.ts`, and the folder README say the two functions are one unit. Docs partial gets the same sentence.

From the notes file:

- Hono and h3 (2, 3): buffer alongside seeding and put the proxy back on the framework request. h3 types `event.req` as readonly, so it goes through a cast.
- h3 env (4): seeds from `event.req.runtime?.cloudflare?.env`.
- Nest headers (5): a guard can set headers on the platform response before throwing, and Nest's exception filter keeps them. Verified on platform-express 11. The short-circuit now carries WWW-Authenticate, CORS, Retry-After, and Set-Cookie. Response phase stays unavailable, as documented.
- Nest `Object.assign` (8): kept flat keys instead of nesting under one property. The partial teaches the same shape on all four frameworks, and nesting would make Nest the odd one out. The guard now refuses a key that already exists on the request or names a property Nest's decorators read (`body`, `query`, `params`, and the rest, from `route-params-factory`). Both of your cases fail with a clear 500 instead of a crash or a silent swap.
- Nest method (6): CONNECT, TRACE, and TRACK map to GET on the Web request.
- Per-route fold (7): `toNestGuard` hoisted to a const in `app.ts` and in the partial's example and agent prompt.
- Hono order: one sentence in the partial.
- Docs link (10): the path shape is right; `introduction` and `installing` are live at the same shape. It 404s until supabase#50514 merges, which follows this PR the same day.

